PR #3, opened 2021-09-03
Added logging capabilities to the _Printer.

facundobatista:
- [X] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -48,14 +48,14 @@ def get_terminal_width():
 def get_log_filepath(appname):
     """Provide a filepath for logging into."""
     # XXX Facundo 2021-09-03: this will change heavily in next couple of branches
-    _, filepath = tempfile.mkstemp(prefix=f"charmcraft-{appname}-log-")
+    _, filepath = tempfile.mkstemp(prefix=f"{appname}-")
     return filepath
 
 
 class _Printer:
     """Handle writing the different messages to the different outputs (out, err and log)."""
 
-    def __init__(self, log_filepath: str):
+    def __init__(self, log_filepath: str) -> None:
         # holder of the previous message
         self.prv_msg: Optional[_MessageInfo] = None
 
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -19,6 +19,7 @@
 import enum
 import shutil
 import sys
+import tempfile
 from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Optional, TextIO, Union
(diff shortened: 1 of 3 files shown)


cmatsuoka (approved):
LGTM.

craft_cli/messages.py
```diff
@@ -44,15 +45,25 @@ def get_terminal_width():
     return shutil.get_terminal_size().columns
 
 
+def get_log_filepath(appname):
+    """Provide a filepath for logging into."""
+    # XXX Facundo 2021-09-03: this will change heavily in next couple of branches
+    _, filepath = tempfile.mkstemp(prefix=f"charmcraft-{appname}-log-")
```
sergiusens, line 51: isn't appname supposed replace charmcraft here?

sergiusens: can we document appname?

craft_cli/messages.py
```diff
@@ -44,15 +45,25 @@ def get_terminal_width():
     return shutil.get_terminal_size().columns
 
 
+def get_log_filepath(appname):
```
sergiusens, line 48:
```suggestion
def get_log_filepath(appname: str):
```

craft_cli/messages.py
```diff
@@ -44,15 +45,25 @@ def get_terminal_width():
...
+    # XXX Facundo 2021-09-03: this will change heavily in next couple of branches
+    _, filepath = tempfile.mkstemp(prefix=f"charmcraft-{appname}-log-")
+    return filepath
+
+
 class _Printer:
     """Handle writing the different messages to the different outputs (out, err and log)."""
 
-    def __init__(self):
+    def __init__(self, log_filepath: str):
```
sergiusens, line 58:
```suggestion
    def __init__(self, log_filepath: str) -> None:
```

craft_cli/messages.py
sergiusens: there are no tests for this method

craft_cli/messages.py
facundobatista: **this will change heavily in next couple of branches**

Added a line about "appname" in the big PR.

craft_cli/messages.py
facundobatista: **this will change heavily in next couple of branches**

Tests exist in the "big PR".

craft_cli/messages.py
facundobatista: Removed "charmcraft" from here, anyway.

craft_cli/messages.py
facundobatista: Nice catch

sergiusens (approved):

PR #5, opened 2021-09-08
Provide a log filepath that complies with our requirements (CRAFT-96).

facundobatista:

-----


craft_cli/messages.py
```diff
@@ -39,23 +41,51 @@ class _MessageInfo:
...
     return shutil.get_terminal_size().columns
 
 
-def get_log_filepath(appname):
-    """Provide a filepath for logging into."""
-    # XXX Facundo 2021-09-03: this will change heavily in next couple of branches
-    _, filepath = tempfile.mkstemp(prefix=f"{appname}-")
-    return filepath
+def get_log_filepath(appname: str) -> pathlib.Path:
+    """Provide a filepath for logging into.
```
sergiusens, line 54:
```suggestion
    """Provide a unique filepath for logging.
```

craft_cli/messages.py
```diff
@@ -39,23 +41,51 @@ class _MessageInfo:
 # the different modes the Emitter can be set
 EmitterMode = enum.Enum("EmitterMode", "QUIET NORMAL VERBOSE TRACE")
 
+# the limit to how many log files to have
+LOG_FILES_LIMIT = 5
+
```
sergiusens, line 46: This renders nicely in sphinx

```suggestion
LOG_FILES_LIMIT = 5
"""The limit to how many log files to have."""
```

I might personally prefer MAX over LIMIT

craft_cli/messages.py
```diff
@@ -39,23 +41,51 @@ class _MessageInfo:
...
-    return filepath
+def get_log_filepath(appname: str) -> pathlib.Path:
+    """Provide a filepath for logging into.
+
+    The app name is used for both the directory where the logs are located and each log name.
+
+    Rules:
+    - use an appdirs provided directory
+    - base filename is <appname>.<timestamp with microseconds>.log
+    - it rotates until it gets to ROTATION_LIMIT
```
sergiusens, line 61: ROTATION_LIMIT is probably the other constant
```suggestion
    - it rotates until it reaches :data:`.LOG_FILES_LIMIT`
```

craft_cli/messages.py
```diff
@@ -39,23 +41,51 @@ class _MessageInfo:
...
+    - after limit is achieved, remove the exceeding files
+    - ignore other non-log files in the directory
+
+    Existing files are not renamed (no need, as each name is unique) nor gzipped (they may
+    be currently in use by other process).
```
sergiusens, line 66:
```suggestion
    be currently in use by another process).
```

craft_cli/messages.py
```diff
@@ -39,23 +41,51 @@ class _MessageInfo:
...
+    """
+    basedir = pathlib.Path(appdirs.user_log_dir()) / appname
+    filename = f"{appname}-{datetime.now():%Y%m%d-%H%M%S.%f}.log"
+
+    # ensure the basedir is there
+    basedir.mkdir(exist_ok=True)
+
+    # check if we doesn't have too many logs in the dir, and remove the exceeding ones (note
```
sergiusens, line 74:
```suggestion
    # check if we have too many logs in the dir, and remove the exceeding ones (note
```

craft_cli/messages.py
cmatsuoka: Agree with Sergio, `MAX_LOG_FILES` could be better here.

craft_cli/messages.py
facundobatista: Improved.

craft_cli/messages.py
facundobatista: Changed to `_MAX_LOG_FILES`, to also make it private (not intended to be part of the public API)

craft_cli/messages.py
facundobatista: Improved. Don't know how to really see this, though.

sergiusens: "make docs" at the root of the project should construct documentation

cmatsuoka (approved):
Thanks Facundo, lgtm.

craft_cli/messages.py
facundobatista: It did, thanks.

craft_cli/messages.py
```diff
@@ -51,30 +51,30 @@ def get_terminal_width() -> int:
...
-    """Provide a filepath for logging into.
+    """Provide a unique filepath for logging.
 
     The app name is used for both the directory where the logs are located and each log name.
 
     Rules:
     - use an appdirs provided directory
     - base filename is <appname>.<timestamp with microseconds>.log
-    - it rotates until it gets to ROTATION_LIMIT
+    - it rotates until it gets to reaches :data:`._MAX_LOG_FILES`
```
sergiusens, line 61: I would make this public as it is documented

craft_cli/messages.py
```diff
@@ -42,7 +42,7 @@ class _MessageInfo:
 EmitterMode = enum.Enum("EmitterMode", "QUIET NORMAL VERBOSE TRACE")
 
 # the limit to how many log files to have
-LOG_FILES_LIMIT = 5
+_MAX_LOG_FILES = 5
```
sergiusens, line 45: I would make this public again and document it so a user of the API is exposed to the default in documentation.

sergiusens (approved):
Looks good, I would prefer for the constant to be public as it is documented.

craft_cli/messages.py
facundobatista: As we talked, made everything private.

craft_cli/messages.py
facundobatista: As we talked, made everything private.

PR #7, opened 2021-09-15
Added the 'progress' method to send progress messages (CRAFT-98).

facundobatista:

-----

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -269,7 +269,7 @@ class Emitter:
     def progress(self, text: str) -> None:
         """Progress information for a multi-step command.
 
-        This is normally to present several separated text messages.
+        This is normally used to present several separated text messages.
 
         These messages will be truncated to the terminal's width, and overwritten by the next
         line (unless verbose/trace mode).
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -33,6 +33,7 @@ class _MessageInfo:
 
     stream: Union[TextIO, None]
     text: str
+    ephemeral: bool = False
     use_timestamp: bool = False
     end_line: bool = False
     created_at: datetime = field(default_factory=datetime.now)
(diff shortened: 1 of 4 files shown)


sergiusens (approved):

craft_cli/messages.py
```diff
@@ -252,6 +265,33 @@ def trace(self, text: str) -> None:
...
+        line (unless verbose/trace mode).
+        """
+        if self.mode == EmitterMode.QUIET:
+            # will not be shown in the screen (always logged to the file)
+            stream = None
+            use_timestamp = False
+            ephemeral = True
+        elif self.mode == EmitterMode.NORMAL:
+            # show the indicated message to stderr (ephemeral) and log it
+            stream = sys.stderr
```
cmatsuoka, line 284: Why are progress messages sent to stderr instead of stdout? Wouldn't that result in unexpected output patterns e.g. when someone runs the application redirecting or piping stdout (and expecting all messages to go to the redirected destination)?

craft_cli/messages.py
```diff
@@ -252,6 +265,33 @@ def trace(self, text: str) -> None:
         stream = sys.stderr if self.mode == EmitterMode.TRACE else None
         self.printer.show(stream, text, use_timestamp=True)  # type: ignore
 
+    @_init_guard
+    def progress(self, text: str) -> None:
+        """Progress information for a multi-step command.
+
+        This is normally to present several separated text messages.
```
cmatsuoka, line 272: Maybe "normally used to"?

craft_cli/messages.py
facundobatista: The idea is: **Everything** is sent to stderr, except for the "expected command output" (see [the table](https://docs.google.com/document/d/1Pe-0ED6db53SmrUGIAgVMzxeCOGJQwsZJWf7jzFSagQ)) so if you want to run `charmcraft` from other process and capture the result, you just send stderr to /dev/null and whatever you got is just the result of the command.

craft_cli/messages.py
facundobatista: nice catch, fixed

cmatsuoka (approved):
+1, thanks for the clarification.

PR #8, opened 2021-09-17
Added the API and infrastructure to show a progress bar (CRAFT-99).

facundobatista:

-----

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -267,6 +267,8 @@ class _Progresser:
 
     def advance(self, amount: Union[int, float]) -> None:
         """Show a progress bar according to the informed advance."""
+        if amount < 0:
+            raise ValueError("The advance amount cannot be negative")
         if self.delta:
             self.accumulated += amount
         else:
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -48,6 +48,9 @@ EmitterMode = enum.Enum("EmitterMode", "QUIET NORMAL VERBOSE TRACE")
 # the limit to how many log files to have
 _MAX_LOG_FILES = 5
 
+# the char used to draw the progress bar ('FULL BLOCK')
+PROGRESS_BAR_SYMBOL = "█"
+
 
 def _get_terminal_width() -> int:
     """Return the number of columns of the terminal."""
@@ -162,7 +165,7 @@ class _Printer:
         # message (truncated, if needed)
         if bar_width > 0:
             completed_width = math.floor(bar_width * min(bar_percentage, 100))
-            completed_bar = "#" * completed_width
+            completed_bar = PROGRESS_BAR_SYMBOL * completed_width
             empty_bar = " " * (bar_width - completed_width)
             line = f"{maybe_cr}{message.text} [{completed_bar}{empty_bar}] {numerical_progress}"
         else:
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -17,23 +17,26 @@
 """Support for all messages, ok or after errors, to screen and log file."""
 
 import enum
+import math
 import pathlib
 import shutil
 import sys
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import Optional, TextIO, Union
+from typing import Literal, Optional, TextIO, Union
 
 import appdirs
 
 
 @dataclass
-class _MessageInfo:
+class _MessageInfo:  # pylint: disable=too-many-instance-attributes
     """Comprehensive information for a message that may go to screen and log."""
 
     stream: Union[TextIO, None]
     text: str
     ephemeral: bool = False
+    bar_progress: Union[int, float, None] = None
+    bar_total: Union[int, float, None] = None
     use_timestamp: bool = False
     end_line: bool = False
     created_at: datetime = field(default_factory=datetime.now)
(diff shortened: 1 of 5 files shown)


craft_cli/messages.py
```diff
@@ -134,14 +137,53 @@ def _write_line(self, message: _MessageInfo) -> None:
...
+        # terminal size minus the text and numerical progress, and 5 (the cursor at the end,
+        # two spaces before and after the bar, and two surrounding brackets)
+        terminal_width = _get_terminal_width()
+        bar_width = terminal_width - len(message.text) - len(numerical_progress) - 5
+
+        # only show the bar with progress if there is enough space, otherwise just the
+        # message (truncated, if needed)
+        if bar_width > 0:
+            completed_width = math.floor(bar_width * min(bar_percentage, 100))
+            completed_bar = "#" * completed_width
```
cmatsuoka, line 165: We could consider changing to a block character at some point in the future.

tests/unit/test_messages_emitter.py
```diff
@@ -274,6 +274,46 @@ def test_progress_in_verboseish_modes(get_initiated_emitter, mode):
...
+        call().show(sys.stderr, "some text", ephemeral=True),
+    ]
+    assert progresser.total == 5000
+    assert progresser.text == "some text"
+    assert progresser.stream == sys.stderr
+    assert progresser.delta is True
+
+
+def test_progressbar_with_delta_false(get_initiated_emitter):
+    """Init _Progresser with delta=Falsee."""
```
cmatsuoka, line 300: Typo, extra "e" in `False`.

tests/unit/test_messages_helpers.py
```diff
@@ -127,3 +129,46 @@ def test_getlogpath_ignore_other_files(test_log_dir, monkeypatch):
...
+
+def test_progresser_absolute_mode():
+    """Just use _Progresser as a context manager in absolute mode."""
+    stream = sys.stdout
+    text = "test text"
+    total = 123
+    fake_printer = MagicMock()
+    with _Progresser(fake_printer, total, text, stream, delta=False) as progresser:
+        progresser.advance(20)
+        progresser.advance(30)
```
cmatsuoka, line 145: I think we could also test for out of range values, and also floating point numbers since we allow floats.

tests/unit/test_messages_helpers.py
```diff
@@ -127,3 +129,46 @@ def test_getlogpath_ignore_other_files(test_log_dir, monkeypatch):
...
+
+def test_progresser_delta_mode():
+    """Just use _Progresser as a context manager in delta mode."""
+    stream = sys.stdout
+    text = "test text"
+    total = 123
+    fake_printer = MagicMock()
+    with _Progresser(fake_printer, total, text, stream, delta=True) as progresser:
+        progresser.advance(20)
+        progresser.advance(30)
```
cmatsuoka, line 161: We could also test for delta values that bring the accumulated total to be out of range, and also using floating point numbers since we allow floats.

tests/unit/test_messages_printer.py
```diff
@@ -232,6 +232,170 @@ def test_writeline_having_previous_message_ephemeral(capsys, monkeypatch, log_fi
...
+    """The indicated progress exceeds the total."""
+    monkeypatch.setattr(messages, "_get_terminal_width", lambda: 40)
+    printer = _Printer(log_filepath)
+
+    msg = _MessageInfo(sys.stdout, "test text", bar_progress=120, bar_total=100)
+    printer._write_bar(msg)
+
+    out, _ = capsys.readouterr()
+    assert len(out) == 39
+    assert out == "test text [###################] 120/100"
```
cmatsuoka, line 292: Not related to this test, but will the output look correct if `bar_progress` is set to something like `1000/100`, and then back to `120/100`?

craft_cli/messages.py
```diff
@@ -134,14 +137,53 @@ def _write_line(self, message: _MessageInfo) -> None:
...
+        elif self.prv_msg.ephemeral:
+            # the last one was ephemeral, overwrite it
+            maybe_cr = "\r"
+        else:
+            # complete the previous line, leaving that message ok
+            maybe_cr = ""
+            print(flush=True, file=self.prv_msg.stream)
+
+        numerical_progress = f"{message.bar_progress}/{message.bar_total}"
+        bar_percentage = min(message.bar_progress / message.bar_total, 1)  # type: ignore
```
cmatsuoka, line 154: Should we clamp out of range progress values in the model or only for display? What is the meaning of negative progress values, if `_Progresser.advance()` is called with a negative number in the absolute case, or if it falls below 0 in the delta case?

craft_cli/messages.py
facundobatista: Yes. @sergiusens already mentioned it. I just didn't know which character to use. Do you have one that we know to be in the different fonts that are used in terminals both in Linux and Windows?

tests/unit/test_messages_emitter.py
facundobatista: Fixed, thanks!

tests/unit/test_messages_helpers.py
facundobatista: Good idea about floating point. Regarding "out of range", we're not doing anything special with them (the code that actually *shows* the progress deals with that)... what kind of test are you suggesting? Just that it doesn't crash?

tests/unit/test_messages_helpers.py
facundobatista: Same question above.

tests/unit/test_messages_printer.py
facundobatista: Yes. Each time the line is drawn, it uses all the terminal's width.

craft_cli/messages.py
facundobatista: We allow out of range values. The bar itself for sure won't reflect it (it will show just full), but the values shown indicate that. My idea here is that we will not "hide" a detail like that in the lib. 

If the app has a bug or something that gets into this case, the user and us would see it, and fix it. If the lib hides it, the progress would just get "stuck at 100%", while the app continues working.

On the other hand, negative values should not be allowed, as they are just wrong. I changed `_Progresser` to raise an error in that case.

craft_cli/messages.py
sergiusens: This should work https://github.com/verigak/progress/blob/master/progress/bar.py#L52, from this post, looks like the ones mentioned in this table https://mike42.me/blog/2018-06-make-better-cli-progress-bars-with-unicode-block-characters would work too

sergiusens (approved):

craft_cli/messages.py
facundobatista: I'd use the one that you mentioned first. It's unicode name is 'FULL BLOCK'... looks simple enough, it should be in all fonts (who know!).

We can leave "sub char resolution" for the future, IMO. I just opened #9 to not forget about it.

facundobatista: Pushed the version using the FULL BLOCK.

tests/unit/test_messages_helpers.py
cmatsuoka: Yes, to make sure it doesn't crash.

cmatsuoka (approved):
Thanks for the clarification.

PR #10, opened 2021-09-24
Added a supervisor to show a spinner and passed time in a long-standing message (CRAFT-100).

facundobatista:

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -100,11 +100,25 @@ def _get_log_filepath(appname: str) -> pathlib.Path:
 
 
 class _Spinner(threading.Thread):
-    """A supervisor thread that will repeat long-standing messages with a spinner besides it."""
+    """A supervisor thread that will repeat long-standing messages with a spinner besides it.
+
+    This will be a long-lived single thread that will supervise each message received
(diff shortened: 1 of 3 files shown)


craft_cli/messages.py
```diff
@@ -89,6 +99,68 @@ def _get_log_filepath(appname: str) -> pathlib.Path:
     return basedir / filename
 
 
+class _Spinner(threading.Thread):
+    """A supervisor thread that will repeat long-standing messages with a spinner besides it."""
+
+    def __init__(self, printer: "_Printer"):
+        super().__init__()
+        # special flag used to stop itself
```
cmatsuoka, line 107: Maybe better as `special flag used to stop the spinner thread`? (otherwise it seems that itself == the flag)

cmatsuoka (approved):
Interesting logic, not immediately obvious but it seems to work well.

craft_cli/messages.py
facundobatista: ah, good idea, thanks!

facundobatista:
Added an explanation on how it works in the class' docstring.

igorljubuncic (approved):

igorljubuncic:
Looks all right.

PR #13, opened 2021-09-29
Added 'open_stream' to the Emitter (CRAFT-101).

facundobatista:

This includes the proper context manager to supervise the third party call and a pipe reader thread which is the one that actually gets the info from the subprocess output streams and sends it to printer.

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -91,7 +91,7 @@ def _get_log_filepath(appname: str) -> pathlib.Path:
     filename = f"{appname}-{datetime.now():%Y%m%d-%H%M%S.%f}.log"
 
     # ensure the basedir is there
-    basedir.mkdir(exist_ok=True)
+    basedir.mkdir(exist_ok=True, parents=True)
 
     # check if we have too many logs in the dir, and remove the exceeding ones (note
     # that the defined limit includes the about-to-be-created file, that's why the "-1")
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -19,8 +19,10 @@
 import enum
 import itertools
 import math
+import os
 import pathlib
 import queue
+import select
 import shutil
 import sys
 import threading
@@ -60,6 +62,9 @@ _SPINNER_THRESHOLD = 2
 
 # seconds between each spinner char
 _SPINNER_DELAY = 0.1
+
+# the size of bytes chunk that the pipe reader will read at once
+_PIPE_READER_CHUNK_SIZE = 4096
 
 
 def _get_terminal_width() -> int:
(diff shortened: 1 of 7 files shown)


craft_cli/messages.py
```diff
@@ -61,6 +63,9 @@ class _MessageInfo:  # pylint: disable=too-many-instance-attributes
 # seconds between each spinner char
 _SPINNER_DELAY = 0.1
 
+# the size of bytes chunk that the pipe reader will read at once
+_PIPE_READER_CHUNK_SIZE = 4096
```
cmatsuoka, line 67: Shouldn't we make it line-oriented instead of chunk-oriented, since this will be used for text and not for binary data?

facundobatista: But you can't.

You have a pipe, and you read bytes from the pipe. You have to read chunks from the pipe (reading byte by byte is too inefficient). And you don't know when in the future you will receive a new line. 

Furthermore, you can not interpret this as text as can not decode the bytes until you're sure you don't have only parts of the bytes; IOW, don't fall in something like the following:

```
>>> "moño".encode("utf8")[:3]
b'mo\xc3'
>>> _.decode("utf8")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc3 in position 2: unexpected end of data
``` 

So it reads chunks of bytes, and then YES, it builds **lines** and those lines are dcoded and sent to the printer (which is line-oriented).

cmatsuoka: Ah yes, there's the encoding problem. We may want to adjust the buffer size however, but let's see how this behaves in a real use case.

craft_cli/messages.py
```diff
@@ -388,6 +393,92 @@ def advance(self, amount: Union[int, float]) -> None:
...
+        self.printer = printer
+        self.stream = stream
+
+    def _write(self, data: bytes) -> None:
+        """Convert the byte stream into unicode lines and send it to the printer."""
+        pointer = 0
+        data = self.remaining_content + data
+        while True:
+            # get the position of next newline (find starts in pointer position)
+            newline_position = data.find(b"\n", pointer)
```
cmatsuoka, line 422: Since data is a byte array containing characters encoded in utf-8, what happens if 0x0a is part of the code for a different unicode character?

facundobatista: It can't. All UTF8 bytes start with `1`, except those unicode points from 0 to 127. See "Encoding" [here](https://en.wikipedia.org/wiki/UTF-8).

cmatsuoka (approved):
Thanks for the clarifications, let's see how this will work in a real use case!

PR #16, opened 2021-10-06
Added error report support (CRAFT-103).

facundobatista:

-----

### DIFF
--- a/craft_cli/errors.py
+++ b/craft_cli/errors.py
@@ -22,23 +22,23 @@ from typing import Optional
 class CraftError(Exception):
     """Signal a program error with a lot of information to report.
 
-    - message: the main message to the user, to be shown as first line (and probably
-      only that, according to the different modes); note that in some cases the log
-      location will be attached to this message.
+    :ivar message: the main message to the user, to be shown as first line (and
+      probably only that, according to the different modes); note that in some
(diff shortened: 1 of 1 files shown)


craft_cli/errors.py
```diff
@@ -0,0 +1,59 @@
...
+    - resolution: an extra line indicating to the user how the error may be fixed or
+      avoided (to be shown together with 'message')
+
+    - docs_url: an URL to point the user to documentation (to be shown together
+      with 'message')
+
+    - reportable: if an error report should be sent to some error-handling backend (like
+      Sentry)
+
+    - retcode: the code to return when the application finishes
```
sergiusens, line 41: Can these be of the form?

```suggestion
    :ivar message: the main message to the user, to be shown as first line (and probably
    only that, according to the different modes); note that in some cases the log
    location will be attached to this message.

    :ivar details: the full error details received from a third party which originated
    the error situation

    :ivar resolution: an extra line indicating to the user how the error may be fixed or
    avoided (to be shown together with 'message')

    :ivar docs_url: an URL to point the user to documentation (to be shown together
    with 'message')

    :ivar reportable: if an error report should be sent to some error-handling backend (like
    Sentry)

    :ivar retcode: the code to return when the application finishes
```

PR #17, opened 2021-10-18
messages: support appdirs user_log_dir for darwin

sergiusens:
appdirs.user_log_dir takes appname as the first parameter and while
the code path works when it is None for windows and linux, it runs
`os.path.expandpath("~/...", appname)` and darwin which fails as
appname is None.

https://github.com/ActiveState/appdirs/blob/48357882e5c095003e10f8351b405bf54e41424f/appdirs.py#L395

Signed-off-by: Sergio Schvezov <sergio.schvezov@canonical.com>

-------

-----

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,7 +44,7 @@ jobs:
   tests:
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


facundobatista (approved):
Wonderful.

PR #23, opened 2021-10-20
Allow a double stop (ok/error), being a noop in that case.

facundobatista:

-----


sergiusens (commented):
Is there a way to start without init?

PR #37, opened 2021-11-15
Removed all pinning infrastructure, now we favor setup.cfg as source of truth.

facundobatista:

-----


sergiusens:
As I read this I do have some conflicts while looking at this, especially once we do branching to support point releases. While it does seem like the right thing to do to fix the library itself if it does not work with it's latest runtime dependencies it feels unnecessary to have to update to the latest flake8 or black, or any other static tooling changes as those, while they improve the product, do not change the consumers interaction with the library. Thoughts on this?

facundobatista:
Let's talk about this in a meeting.

sergiusens:
Closing in favour of #39

PR #38, opened 2021-11-16
Declare public API names

cjwatson:
`mypy --strict` enforces these by default.

I'm not sure whether `Emitter` should be public.  It doesn't have an underscore, but it also isn't used by charmcraft - I guess it might be needed to implement test fixtures?  I decided to err on the side of strictness and you can always open it up if you want.

- [X] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?  (Canonical employee, so not required.)

-----

### DIFF
--- a/craft_cli/errors.py
+++ b/craft_cli/errors.py
@@ -16,6 +16,10 @@
 
 """Error classes."""
 
+__all__ = [
+    "CraftError",
+]
+
 from typing import Optional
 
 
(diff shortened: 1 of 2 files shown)


facundobatista:
Hola Colin,

We have mixed feelings about the "publicness" of `Emitter`. As a name, it's not used from outside (so the PR here is fine, will land it). But it's the "main interface with all the message system", so it's weird to declare it "private".

Thanks for this work!

facundobatista (approved):
Thanks!

PR #41, opened 2021-11-29
Implement comparison for CraftError

### LINKED ISSUE #40: Implement `__eq__` for `CraftError`
For testing purposes it would be nice to be able to compare an expected instance of `CraftError` with a given one.

e.g.

```
from craft_cli import CraftError

expected = CraftError("foo")

given = CraftError("foo")

# currently

expected == given
False
```

jugmac00:
This is especially helpful for testing purposes so that the actual instance
of CraftError can be compared with the expected one.

This fixes #40

(not necessary -> Canonical employee)

-----

### DIFF
--- a/craft_cli/errors.py
+++ b/craft_cli/errors.py
@@ -62,13 +62,12 @@ class CraftError(Exception):
         if isinstance(other, CraftError):
             return all(
                 [
-                    self.args == getattr(other, "args"),
-                    self.details == getattr(other, "details"),
-                    self.resolution == getattr(other, "resolution"),
-                    self.docs_url == getattr(other, "docs_url"),
-                    self.reportable == getattr(other, "reportable"),
-                    self.retcode == getattr(other, "retcode"),
+                    self.args == other.args,
+                    self.details == other.details,
+                    self.resolution == other.resolution,
+                    self.docs_url == other.docs_url,
+                    self.reportable == other.reportable,
+                    self.retcode == other.retcode,
                 ]
             )
-        else:
-            return NotImplemented
+        return NotImplemented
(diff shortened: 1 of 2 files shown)


craft_cli/errors.py
```diff
@@ -57,3 +57,18 @@ def __init__(
         self.docs_url = docs_url
         self.reportable = reportable
         self.retcode = retcode
+
+    def __eq__(self, other):
+        if isinstance(other, CraftError):
+            return all(
+                [
+                    self.args == getattr(other, "args"),
```
facundobatista, line 65: Why not doing `other.args` directly?

tests/unit/test_errors.py
```diff
@@ -0,0 +1,22 @@
+from craft_cli.errors import CraftError
+
+
+def test_crafterror_is_comparable():
```
facundobatista, line 4: Please add two tests to check for equal and different for the rest of `CraftError` attributes (two tests is enough, each parametrized to test the different attributes).

Thanks!

facundobatista (changes requested):
Hello! Great work! Thanks! 

I annotated a couple of details in line, thanks again.

craft_cli/errors.py
jugmac00: Good catch. This is a leftover before I did the `isinstance` check -> fixed

tests/unit/test_errors.py
jugmac00: I created two tests - I hope they are the way you intended.

facundobatista (approved):
Yes, awesome, exactly that, thanks!!

PR #46, opened 2021-12-18
New functionality from charmcraft (CRAFT-706).

facundobatista:

-----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ test-pydocstyle:
 
 .PHONY: test-pylint
 test-pylint:
-	pylint craft_cli --disable=cyclic-import
+	pylint craft_cli
 	pylint tests --disable=missing-module-docstring,missing-function-docstring,redefined-outer-name,protected-access,duplicate-code
 
 .PHONY: test-pyright
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ test-pydocstyle:
 
 .PHONY: test-pylint
 test-pylint:
-	pylint craft_cli
+	pylint craft_cli --disable=cyclic-import
 	pylint tests --disable=missing-module-docstring,missing-function-docstring,redefined-outer-name,protected-access,duplicate-code
 
 .PHONY: test-pyright
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -1,0 +1,366 @@
+# Copyright 2021 Canonical Ltd.
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License version 3 as published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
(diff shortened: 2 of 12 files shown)


craft_cli/dispatcher.py
```diff
@@ -0,0 +1,359 @@
...
+    It also must/can override some methods for the proper command behaviour (see each
+    method's docstring).
+
+    The subclass must be declared in the corresponding section of main.COMMAND_GROUPS,
+    and will receive and store this group on instantiation (if overriding `__init__`, the
+    subclass must pass it through upwards).
+    """
+
+    common = False
+    needs_config = False
```
facundobatista, line 90: I think I will remove this attribute from here.

If the commands needs a config or not is something very charmcraft specific, and it actually cannot be automatically verified here (charmcraft needs to check it!).

craft_cli/helptexts.py
```diff
@@ -0,0 +1,291 @@
...
+import textwrap
+from operator import attrgetter
+from typing import TYPE_CHECKING, List, Tuple
+
+if TYPE_CHECKING:
+    from craft_cli.dispatcher import BaseCommand, CommandGroup
+
+
+# max columns used in the terminal
+TERMINAL_WIDTH = 72
```
cmatsuoka, line 27: Why do we use 72 instead of, say, 78?

Makefile
```diff
@@ -91,7 +91,7 @@ test-pydocstyle:
 
 .PHONY: test-pylint
 test-pylint:
-	pylint craft_cli
+	pylint craft_cli --disable=cyclic-import
```
cmatsuoka, line 94: Why is this needed? Is it reporting a false positive?

cmatsuoka (approved):
Since this was extracted from Charmcraft I assume it will provide all the functionality needed by this tool. We'll see how generic it is when integrating with other craft tools, but generally it LGTM. We may need to rethink the legacy fallback mechanism implementation (as previously discussed) because the spike just parsed arguments by hand.

craft_cli/dispatcher.py
```diff
@@ -0,0 +1,359 @@
+# Copyright 2021 Canonical Ltd.
```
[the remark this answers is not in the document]
facundobatista: Updated copyright years here and in all the files.

craft_cli/helptexts.py
```diff
@@ -0,0 +1,291 @@
+# Copyright 2021 Canonical Ltd.
```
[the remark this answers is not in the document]
facundobatista: Yes.

craft_cli/helptexts.py
facundobatista: I wanted to be conservative here. Note that this module does not present the texts to the user, just builds them and returns them. Maybe the caller would want to indent them one or two columns... so I decided not to use "80", and choose what PEP8 recommends for docstrings, which is "72".

But I'm to change it if you have reasons.

Makefile
facundobatista: Yes. Note that import order in `craft_cli/__init__.py` is defined to avoid the cyclic import. But pylint falls in "the trap" anyway...

craft_cli/helptexts.py
cmatsuoka: I'm good with 72, I was just curious about this exact value.

Makefile
sergiusens: If you need to toggle a check, do it here instead https://github.com/canonical/craft-cli/blob/main/.pylintrc#L72

sergiusens (changes requested):
just wanted to flag that the check should be blocked in configuration for all IDEs to benefit

Makefile
facundobatista: ack, improved.

sergiusens (approved):

PR #47, opened 2022-01-05
Fix tox environment setup

cjwatson:
The requirements files were removed a while ago, but `tox.ini` still
referenced them.


jugmac00 (approved):
LGTM

@maintainers In the long run it would be great to have something similar as in craft-providers, ie run tox in CI, so the local dev environment cannot break, see https://github.com/canonical/craft-providers/pull/86

sergiusens (approved):

PR #48, opened 2022-01-05
Only show spinners if the stream is a terminal

cjwatson:
If the output stream isn't a terminal (for example, if *craft is running
in a Launchpad build job), then showing spinners isn't very useful and
results in build logs parts of which are quite hard to read.  Limit this
behaviour to terminals.

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -37,6 +37,7 @@ import time
 import traceback
 from dataclasses import dataclass, field
 from datetime import datetime
+from functools import lru_cache
 from typing import Literal, Optional, TextIO, Union
 
 import platformdirs
@@ -51,6 +52,11 @@ except ImportError:
 from craft_cli import errors
 
 
+@lru_cache
+def _stream_is_terminal(stream: Union[TextIO, None]) -> bool:
+    return getattr(stream, "isatty", lambda: False)()
+
+
 @dataclass
 class _MessageInfo:  # pylint: disable=too-many-instance-attributes
     """Comprehensive information for a message that may go to screen and log."""
@@ -63,11 +69,6 @@ class _MessageInfo:  # pylint: disable=too-many-instance-attributes
     use_timestamp: bool = False
     end_line: bool = False
     created_at: datetime = field(default_factory=datetime.now)
-
-    @property
-    def stream_is_terminal(self) -> bool:
-        """Return whether the message's stream is a terminal."""
-        return getattr(self.stream, "isatty", lambda: False)()
 
 
 # the different modes the Emitter can be set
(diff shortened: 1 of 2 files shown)


sergiusens (approved):

facundobatista (changes requested):
Thanks for this. Yes, it's a very good idea to avoid the spinner if the output is not a terminal.

That said, two details:

- I'd isolate the "is a terminal" verification to a different function (maybe `is_output_a_terminal` or a better name), as we may use this functionality in the future in other parts of `craft_cli`.

- if we don't want to a spinner, we should avoid starting the spinner's thread (not avoid sending it messages, which may happen from multiple places); there's a part of the code that actually does this in test mode, so that line should just be improved, like:
```
        if not TESTMODE or not output_is_terminal():
            self.spinner.start()
```

cjwatson:
I tried to implement these, but ran into problems, and would appreciate your advice on how you'd like to handle them.

The main difficulty is that there isn't a single output stream: messages have their own `stream` attribute.  In practice, if the stream isn't `None`, I think it's currently only ever `sys.stdout` or `sys.stderr`; but that's still enough to open the possibility of (for example) only stderr being redirected to a file and not stdout, and perhaps this code might want to support other streams in future.  This makes it hard to avoid starting the spinner's thread, because the code that does that doesn't know whether the spinner might be needed for *some* output stream in future.

Having a single `output_is_terminal` kind of function is difficult for similar reasons, but I agree that it would be useful to abstract this a bit, so I moved it into a `stream_is_terminal` property on the message, which I hope is at least a bit of an improvement.

facundobatista:
Hello @cjwatson, thanks for this analysis, and sorry for the delayed response. 

Yes, you're right, we currently have multiple streams, and we may haver others in the future. And we cannot predict if the system will run under the situation of one stream being a terminal (we want the spinner!) and other that is not (we don't want it there).

Because of that, we cannot avoid having all the spinner machinery running, and we need to decide if to show it or not according to the stream used. We're coming back to the original request/idea: "Only show spinners if the stream is a terminal".

Luckily (well, by design) there is one single point of entry for `_Printer` to show the spinner, and that is its `spin` method. And this receives the `message`, which includes to which stream it will be sent.

So, we could do something like the following:

```
    def spin(self, message: _MessageInfo, spintext: str) -> None:
        """Write a line message including a spin text."""
        if stream_is_terminal(message.stream):
            self._write_line(message, spintext=spintext)
```
We may even add a `lru_cache` decorator to `stream_is_terminal`, as it will be called a zillion times but the response will always be the same for the same stream.

What do you think?

Thanks!!

cjwatson:
@facundobatista Hmm, OK.  How's this?  Adding the cache meant that (a) I had to move it out to a top-level function because otherwise the cache would have been per-message, and (b) I had to add a bit more test machinery to avoid test isolation issues, but otherwise it seemed straightforward enough.

facundobatista (approved):
Loved it, thanks! You rock!

PR #53, opened 2022-01-21
Include a section in README about setting up developer environment.

facundobatista:
Also improved a detail in a test because newer pylint complains.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,21 @@ The library provides two main functionalities:
 The [documentation](https://craft-cli.readthedocs.io) is available in Read The Docs.
 
 
+# Setting up the environment
+
+Install at system level:
+
+    sudo snap install pyright
+
+Create a virtual environment, activate it, and install developer dependencies:
+
+    python3 -m venv env
+    source env/bin/activate
+    pip install .[dev]
+
+That's all.
+
+
 # Contributing
 
 A `Makefile` is provided for easy interaction with the project. To see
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,6 +15,21 @@ The library provides two main functionalities:
...
 
+# Setting up the environment
+
+Install at system level:
+
+    sudo snap install pyright
+
+Create a virtual environment, activate it, and install developer dependencies:
+
+    python3 -m venv env
```
cmatsuoka, line 26: Should we use a user-selectable name such as `python3 -m venv <dir name>`?

cmatsuoka (approved):

README.md
```diff
@@ -15,6 +15,21 @@ The library provides two main functionalities:
...
+    source env/bin/activate
+    pip install .[dev]
+
+That's all.
+
+
```
jugmac00, line 32: Just a note.

You could also setup the development environment via

```
tox --devenv <dir name>
```

But right - I just saw you install `tox` the other way round with `.[dev]` - so my suggestion does not make much sense for this setup :-)

README.md
facundobatista: In this case, I prefer to give instructions that you can copy and execute. Not that you need to understand how to modify to change them.

README.md
facundobatista: Yes, there are multiple ways. We just need to indicate one so we enable newcomers to start easily.

PR #54, opened 2022-01-21
Release 0.1.

facundobatista:
Including several files and changes to make releases work.

### DIFF
--- a/HOWTO_RELEASE.md
+++ b/HOWTO_RELEASE.md
@@ -1,0 +1,106 @@
+# Steps
+
+Release early, release often. Don't be lazy.
+
+To use this doc: just replace 0.1.0 with the major.minor.patch version of
+the release. The sequence of commands below should be good to copy and
+paste, but do please pay attention to details!
+
+
+## Preparation
+
+- update main and run tests to verify all is green
+
+    git fetch upstream
+    git merge upstream/main
+    make tests
+
+
+### if it's a minor release (Z == 0)
+
+- tag `main` with only the minor version:
+
+    git tag 0.1
+    git push --tags
+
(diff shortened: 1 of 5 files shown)


HOWTO_RELEASE.md
```diff
@@ -0,0 +1,106 @@
+# Steps
+
+Release early, release often. Don't be lazy.
+
+To use this doc: just replace 0.1.0 with the major.minor.patch version of
+the release. The sequence of commands below should be good to copy and
+paste, but do please pay attention to details!
```
cmatsuoka, line 7: I believe this was meant to be "please do" instead of "do please"?

HOWTO_RELEASE.md
```diff
@@ -0,0 +1,106 @@
...
+- cherry pick the needed commits from main:
+
+   git cherry-pick -m 1 COMMIT-HASH
+   ...
+
+- create release notes from the selected commits
+
+    git log
+
+- tag the release (using those release notes)
```
cmatsuoka, line 55: How do I tag using release notes?

cmatsuoka (approved):
Thanks, it's good to have a document with release information. I must add that to craft-parts as well.

General structure looks good, I just added a couple of notes to verify before merging.

HOWTO_RELEASE.md
facundobatista: As indicated in the next command (it will open you an editor).

HOWTO_RELEASE.md
facundobatista: Just removed the `do`. And also replaced the numbers for generic X.Y.Z (was a leftover).

PR #62, opened 2022-02-04
Restore old examples functionality.

facundobatista:
I broke this when doing tests for first release.


sergiusens:
Maybe add running this as a new step here https://github.com/canonical/craft-cli/blob/main/.github/workflows/tests.yaml#L58 to ensure it at least does not fail to run.

cmatsuoka (approved):

PR #63, opened 2022-02-15
Added Python 3.10 to the tests matrix (CRAFT-742).

facundobatista:


jugmac00:
I am not sure about your intentions to keep Python versions in sync over all configuration files:

`tox.ini` only lists Python 3.8, and `setup.cfg` only lists 3.8 and 3.9.

Usually, I use a tool called [check-python-versions](https://github.com/mgedmin/check-python-versions) as a linter to make sure all files are aligned, but looks like that tool does not fully support gha yet:

```
$ check-python-versions 
setup.py says: 3.8, 3.9
tox.ini says:  3.8

mismatch!
```

jugmac00:
> No worries about `tox`, we will remove it in the future...

Oh, that is a pity. I always found that `tox` makes it very easy, especially as a drive-by contributor,  to setup a development environment, build the project and run the tests - and additionally run the tests for different Python versions, before pushing the changes to the CI, and because of a possibly minor detail all maintainers get notified by an email.

Most open source projects I contribute to run `tox` both locally and via CI, so there is no need to keep configuration files in sync.

I may be biased, though, as I am one of the `tox` maintainers :-)

sergiusens:
On Tue, Feb 15, 2022 at 12:29 PM Jürgen Gmach ***@***.***>
wrote:

> No worries about tox, we will remove it in the future...
>
> Oh, that is a pity. I always found that tox makes it very easy,
> especially as a drive-by contributor, to setup a development environment,
> build the project and run the tests - and additionally run the tests for
> different Python versions, before pushing the changes to the CI, and
> because of a possibly minor detail all maintainers get notified by an email.
>
> Most open source projects I contribute to run tox both locally and via
> CI, so there is no need to keep configuration files in sync.
>
> I may be biased, though, as I am one of the tox maintainers :-)
>

Interesting pitch for tox :-)

I personally have a personal preference for Makefiles due to it's very
simple nature and really don't understand what tox adds aside from
isolation. The github actions thing indeed is a bit of a problem to set up
and it feels we are supporting it for the sake of supporting it but not
really using it in this team.

If we keep tox, could we remove all Makefiles and have neat GHA
integration? Can we get everything ported over from our makefiles? Sphinx'
init auto sets up a Makefile too, is that replaceable?

sergiusens (approved):

jugmac00:
> If we keep tox, could we remove all Makefiles and have neat GHA integration? Can we get everything ported over from our makefiles? Sphinx' init auto sets up a Makefile too, is that replaceable?

Generally, always go with what you and your team is comfortable, but yes, you could replace all Makefiles with `tox` environments.

e.g. in order to create documentation:

```
[testenv:docs]
description = generate docs in HTML format
# align with Python version on Read the Docs
basepython = python3  
deps = -r doc/requirements.txt
commands = sphinx-build -b html docs docs/_build/html
```

You can do similar things for running tests and creating a release.

Let's leave linters for a moment.

The very simplest GHA integration to run the tests for a specific Python version then would be...

Inside the GHA configuration file...
- create a Python matrix
- install `tox`
- run `tox -e py`

This command executes the tests for the Python version `tox` is executed with - e.g. GHA provides Python 3.9, installs `tox` and then runs the equivalent of `tox -e py39`.

There is one problem left with this approach - your other environments are not run, e.g. creating docs, running linters... you could install and run them separately, but then again you need to sync the linters and versions between GHA configuration and your local development environment, either Makefiles or `tox`.

So, for the best integration of `tox` with GHA there is a plugin called https://github.com/ymyzk/tox-gh-actions

Again, in your GHA configuration file...
- create a Python matrix
- install both `tox` and `tox-gh-actions`
- run `python -m tox`

In your `tox.ini` you need to add mapping between `tox` environments and GHA Python versions, e.g.:

```
[gh-actions]
python =
    3.6: py36
    3.7: py37
    3.8: py38, mypy
    3.9: py39, coverage
    3.10: py310, lint
```

On GHA, this would run tests for Python 3.6, tests for Python 3.7, tests for Python 3.8 and mypy, tests for Python 3.9 and create a coverage report, and for Python 3.10 run tests and the linters.

You would only have to touch the GHA configuration file when you add a new Python version - you would never have to touch it to add a new linter or a new version of a linter.

That is the workflow many open source projects have been using for a couple of years (back then with Travis).

Since quite some time this workflow has been improved by using a tool called https://pre-commit.com/

Instead of managing all your different linters via separate `tox` environments, now `pre-commit` is additionally used, which then drives e.g. flake8, black, isort...

e.g. in your `tox.ini` you would add then...

```
[testenv:lint]
description = run all linters
deps = pre-commit
commands = pre-commit run --all-files --show-diff-on-failure
```

The main advantages:
- `pre-commit` can be used as a hook for git - so when you do a `git commit` automatically the linters are run - so you cannot commit badly formatted code (This is pretty awesome).
- you have all linters in one nice configuration file, including version pinning! (no more works on my machine, but fails on CI as a newer linter version was used there)
- you can update all linters with a single command, e.g. once a week, or you could even let this be done by a GitHub bot which creates a pull request each week with new linter verions

That is a lot of stuff :-)

There is a lengthy blog post about using `tox` and GHA by Hynek:
https://hynek.me/articles/python-github-actions/

I gave a 5 min Lightning Talk about running linters locally and on CI:
https://www.youtube.com/watch?v=OnM3KuE7MQM

You could have a look at the following configuration files:

https://github.com/jugmac00/flask-reuploaded/blob/44e35011b4f1c9d8413275a0b2d75a3d257e8efd/tox.ini#L57-L63

https://github.com/jugmac00/flask-reuploaded/blob/44e35011b4f1c9d8413275a0b2d75a3d257e8efd/.pre-commit-config.yaml#L1-L5

https://github.com/jugmac00/flask-reuploaded/blob/44e35011b4f1c9d8413275a0b2d75a3d257e8efd/.github/workflows/main.yml#L20-L39

You can always ping me on Mattermost or set up a hangout.

I could also create a basic pull request to show you how this would look - but I'd do this outside my working hours :-)

facundobatista:
We should move this tox-related conversation to an issue.

PR #69, opened 2022-03-18
Add a 'pause' method to the Emitter (CRAFT-907).

facundobatista:
While paused, the Emitter frees completely the terminal and does not do anything with it. Not even getting logs from it.

Examples 21 and 22 added to easily test this new feature.


cmatsuoka (approved):
Simple and effective, +1.

PR #71, opened 2022-03-22
Improvements/fixes for the pause situation.

facundobatista:
- the _Printer now opens the log in "append" mode; this is particularly so after resuming the previous log is not truncated, but also is an improvement in general, as we allow users of the library to have better control if they provide the log filepath (if they want a fresh start, they can truncate it, before we were not leaving a choice)

- ensure the _Emitter is not used when paused... but I fixed this in a more generic way, ensuring that the _Emitter is not used when stopped! Except for the `ended_ok` and `error` methods, where we provide the possibility of double-endings (ignoring the call), to support race conditions when closing the lib.

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -229,7 +229,7 @@ class _Printer:
         self.prv_msg: Optional[_MessageInfo] = None
 
         # open the log file (will be closed explicitly later)
-        self.log = open(log_filepath, "wt", encoding="utf8")  # pylint: disable=consider-using-with
+        self.log = open(log_filepath, "at", encoding="utf8")  # pylint: disable=consider-using-with
 
         # keep account of output streams with unfinished lines
         self.unfinished_stream: Optional[TextIO] = None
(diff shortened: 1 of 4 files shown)


craft_cli/messages.py
```diff
@@ -602,15 +602,27 @@ def emit(self, record: logging.LogRecord) -> None:
...
-def _init_guard(wrapped_func):
-    """Decorate Emitter methods to be called *after* init."""
+def _active_guard(ignore_when_stopped=False):
+    """Decorate Emitter methods to be called when active.
 
-    def func(self, *args, **kwargs):
-        if not self._initiated:  # pylint: disable=protected-access
-            raise RuntimeError("Emitter needs to be initiated first")
-        return wrapped_func(self, *args, **kwargs)
+    It will check that the emitter is initted and that is not stopped (except when
```
cmatsuoka, line 608: Maybe "initiated" or "initialized"?

tests/unit/test_messages_emitter.py
```diff
@@ -524,6 +538,15 @@ def test_paused_resumed_error(get_initiated_emitter, tmp_path):
     ]
 
 
+def test_paused_cant_show(get_initiated_emitter, tmp_path):
+    """The Emitter cannot show messages when paused."""
+    emitter = get_initiated_emitter(EmitterMode.QUIET)
+
+    with emitter.pause():
+        with pytest.raises(RuntimeError):
```
cmatsuoka, line 546: Good, we need some kind of error instead of failing silently.

cmatsuoka (approved):
Thanks for noticing and fixing this issue.

craft_cli/messages.py
facundobatista: yes, thanks!

PR #72, opened 2022-03-31
Introduce the concept of hidden commands.

facundobatista:


craft_cli/dispatcher.py
```diff
@@ -93,6 +93,9 @@ class BaseCommand:
     - common: if it's a common/starter command, which are prioritized in the help (default to
       False)
 
+    - hidden: do not show in help texts, useful for aliases or depreacated commands (default
```
jugmac00, line 96:
```suggestion
    - hidden: do not show in help texts, useful for aliases or deprecated commands (default
```

sergiusens (approved):
looks good, seems there are new errors picked up by a newer pylint

craft_cli/dispatcher.py
```diff
@@ -101,6 +104,7 @@ class BaseCommand:
     """
 
     common = False
+    hidden = False
```
jugmac00, line 107: As `common` and `hidden` are mutually exclusive would an enum be a good replacement for the two boolean values?

jugmac00 (commented):
There is one typo and I have one question of you do not mind.

jugmac00:
> looks good, seems there are new errors picked up by a newer pylint

In order to prevent random and unrelated linting failures, pinning the linters could be a good idea.

For a couple of years, https://pre-commit.com/ has been widely used predominantly in the python ecosystem, though it is language agnostic. Besides pinning it has many advantages. If you are interested in a quick introduction, please ping me on Mattermost.

craft_cli/dispatcher.py
facundobatista: Thanks!

craft_cli/dispatcher.py
facundobatista: I like "the rule of three". Two cases are too few to start generalizing.

facundobatista:
> > looks good, seems there are new errors picked up by a newer pylint
> 
> In order to prevent random and unrelated linting failures, pinning the linters could be a good idea.
> 
> For a couple of years, https://pre-commit.com/ has been widely used predominantly in the python ecosystem, though it is language agnostic. Besides pinning it has many advantages. If you are interested in a quick introduction, please ping me on Mattermost.

They are not random lint failures, they are failures because the tool is newer than what we tested locally.

We do NOT want to pin the development tools, otherwise we always get stuck in the past. There's no problem in having these failures, I now update the dev tools in my machine, fix these new issues, and the code evolved correctly and we all use latest tools.

jugmac00:
I really appreciate that you took the time to answer my question on enum.

Just a quick follow-up on `pre-commit`.

> We do NOT want to pin the development tools, otherwise we always get stuck in the past.

That is not true for pre-commit. There is an optional and free github service which creates a pull request each week and updates your linters and again optionally it would even fix the possible new linter errors (for e.g. black).

Even if you do not plan to use this service, you can e.g. every time you work on a project or e.g. once a week run `pre-commit autoupdate` and update all pinned linters with a single command in a controlled way.

My reasoning to always pin my linters:
Imagine there is an urgent security or bug fix necessary, and then CI fails for some (not random) but completely unrelated reasons. Then you additionally need to figure out how to fix the linter errors or dig in past build logs and find a linter version which did not cause an error.

But certainly both ways work.

PR #76, opened 2022-04-19
Added init_emitter and emitter fixtures (CRAFT-994).

### LINKED ISSUE #57: Provide a pytest fixture to simplify tests that use `emit`
The functionality is already in charmcraft, need to move it here, but needs to be in a way that installing `craft-cli` makes the plugin available.

facundobatista:
Both are brought from Charmcraft, but here are also tested (as they are not really *used*, but are provided for users of the library).

As these fixtures also work for the project itself (because for development the project itself is installed), I removed the redundant fixture in `conftest.py`. Related to this, instead of enabling that fixture in a couple of test files, in consideration that the "real" `init_emitter` fixture has `autouse=True`, I just turned it off where we didn't want to mess with tests.

Fixes #57.

### DIFF
--- a/craft_cli/pytest_plugin.py
+++ b/craft_cli/pytest_plugin.py
@@ -28,7 +28,7 @@ from craft_cli import messages
 
 
 @pytest.fixture(autouse=True)
-def init_emitter():
+def init_emitter(monkeypatch):
     """Ensure emit is always clean, and initted (in test mode).
 
     Note that the `init` is done in the current instance that all modules already
@@ -42,7 +42,7 @@ def init_emitter():
     os.close(temp_fd)
     temp_logfile = pathlib.Path(temp_logfile)
 
-    messages.TESTMODE = True
+    monkeypatch.setattr(messages, "TESTMODE", True)
     messages.emit.init(
         messages.EmitterMode.QUIET, "test-emitter", "Hello world", log_filepath=temp_logfile
     )
--- a/craft_cli/pytest_plugin.py
+++ b/craft_cli/pytest_plugin.py
@@ -52,7 +52,7 @@ def init_emitter(monkeypatch):
     temp_logfile.unlink()
 
 
-class RegexComparingText(str):
+class _RegexComparingText(str):
     """A string that compares for equality using regex.match."""
 
     def __eq__(self, other):
@@ -62,7 +62,7 @@ class RegexComparingText(str):
         return str.__hash__(self)
 
 
-class RecordingEmitter:
+class _RecordingEmitter:
     """Record what is shown using the emitter and provide a nice API for tests."""
 
     def __init__(self):
@@ -85,7 +85,7 @@ class RecordingEmitter:
     def _check(self, expected_text, method_name, regex, **kwargs):
         """Really verify messages."""
         if regex:
-            expected_text = RegexComparingText(expected_text)
+            expected_text = _RegexComparingText(expected_text)
         expected_call = call(method_name, expected_text, **kwargs)
         for stored_call in self.interactions:
             if stored_call == expected_call:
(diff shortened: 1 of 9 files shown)


craft_cli/pytest_plugin.py
```diff
@@ -0,0 +1,194 @@
...
+    """A string that compares for equality using regex.match."""
+
+    def __eq__(self, other):
+        return bool(re.match(self, other, re.DOTALL))
+
+    def __hash__(self):
+        return str.__hash__(self)
+
+
+class _RecordingEmitter:
```
sergiusens, line 65: why not make this public if referenced in the documentation?

docs/howtos.rst
```diff
@@ -166,3 +166,24 @@ To be able to run another application (in other process) without interfering in
...
+The library provides two fixtures that simplifies the testing of code using the Emitter when 
+using ``pytest``.
+
+One of the fixtures (``init_emitter``) is even set with ``autouse=True``, so it will 
+automatically initialize the Emitter and tear it down after each test. This way there is 
+nothing special you need to do in your code when testing it, just use it.
+
+The other fixture (``emitter``) is very useful to test code interaction with Emitter. It provides an internal recording emitter that has several methods which help to test its usage.
+
+The following example shows a simple usage, please refer to `its reference <craft_cli.dispatcher.html#craft_cli.pytest_plugin._RecordingEmitter>`_ for more information about the provided functionality::
```
sergiusens, line 183: /home/sergiusens/repo/canonical/craft-cli/docs/howtos.rst:3: WARNING: Duplicate explicit target name: "its reference".

I would align this and the previous paragraphs to make it consistent with the first two (line breaks). RecodringEmitter needs to be public for this to render documentation.

```suggestion
The following example shows a simple usage, refer to :class:`craft_cli.pytest_plugin.RecordingEmitter` for more information about the provided functionality::
```

sergiusens: You should probably replace all other instances of this, like 

    :class:`craft_cli.dispatcher.GlobalArgument` 

which had the original conflict.

tests/unit/test_messages_printer.py
```diff
@@ -27,6 +27,11 @@
 from craft_cli.messages import _MessageInfo, _Printer, _Spinner
 
 
+@pytest.fixture(autouse=True)
+def init_emitter():
+    """Disable the automatic init emitter fixture for all this module."""
```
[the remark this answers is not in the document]
facundobatista: ok

tests/unit/test_messages_emitter.py
```diff
@@ -29,6 +29,11 @@
 FAKE_LOG_NAME = "fakelog.log"
 
 
+@pytest.fixture(autouse=True)
+def init_emitter():
+    """Disable the automatic init emitter fixture for all this module."""
```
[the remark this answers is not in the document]
facundobatista: ok

craft_cli/pytest_plugin.py
```diff
@@ -0,0 +1,194 @@
...
+def init_emitter(monkeypatch):
+    """Ensure emit is always clean, and initted (in test mode).
+
+    Note that the `init` is done in the current instance that all modules already
+    acquired.
+    """
+    # init with a custom log filepath so user directories are not involved here; note that
+    # we're not using pytest's standard tmp_path as Emitter would write logs there, and in
+    # effect we would be polluting that temporary directory (potentially messing with
+    # tests, that may need that empty), so we use another one.
```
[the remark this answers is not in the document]
facundobatista: I never end them with a dot, removed it there

craft_cli/pytest_plugin.py
facundobatista: Because it's not meant to be used directly. But if we need it public for the documentation, I'll make it public and annotate in its docstring that is not meant to be used directly...

docs/howtos.rst
facundobatista: I just made all "long lines", as the rest of the text. It leads to more simple diffs when something change in a line (not having to re-work all cut lines).

Nice trick the `:class:` one! Improved the rest of the text.

docs/howtos.rst
```diff
@@ -125,7 +125,7 @@ Define and use other global arguments
 
 To define more automatic global arguments than the ones provided automatically by ``Dispatcher`` (see :ref:`this explanation <expl_global_args>` for more information), use the ``GlobalArgument`` object to create all you need and pass them to the ``Dispatcher`` at instatiaton time.
 
-Check `its reference <craft_cli.dispatcher.html#craft_cli.dispatcher.GlobalArgument>`_ for more information about the parameters needed, but it's very straightforward to create these objects. E.g.::
+Check :class:`its reference <craft_cli.dispatcher.GlobalArgument>` for more information about the parameters needed, but it's very straightforward to create these objects. E.g.::
```
sergiusens, line 128: I would directly expose the class name here:

```suggestion
Check :class:`craft_cli.dispatcher.GlobalArgument` for more information about the parameters needed, but it's very straightforward to create these objects. E.g.::
```

sergiusens: If not, use as you used Dispatcher below

sergiusens (approved):
some take it or leave it comments, thanks for the work

docs/howtos.rst
facundobatista: Improved.

PR #82, opened 2022-06-15
Allow error reporting to not expose the log filepath.

facundobatista:


.pylintrc
```diff
@@ -69,85 +69,14 @@ confidence=
...
-        unpacking-in-except,
-        old-raise-syntax,
-        backtick,
-        long-suffix,
-        old-ne-operator,
-        old-octal-literal,
-        import-star-module-level,
-        non-ascii-bytes-literal,
-        raw-checker-failed,
+disable=raw-checker-failed,
```
cmatsuoka, line 72: Good move, it's cleaner if we disable only what we actually need to disable.

cmatsuoka (approved):
Looks good!

.pylintrc
facundobatista: this was actually forced by new pylint, as they do not support those "disables" anymore (because they dropped python2 support)

PR #87, opened 2022-06-20
Support a global argument that does not have a short form option (CRAFT-1132).

facundobatista:


tests/unit/test_help.py
```diff
@@ -806,3 +807,31 @@ def test_tool_exec_help_command_all():
         "-t, --trace",
         "-v, --verbose",
     ]
+
+
+def test_tool_exec_help_when_globalarg_without_short_form(monkeypatch):
+    """Validate that the args for help are ok without a short form."""
+    new_global = GlobalArgument(
```
sergiusens, line 814: this is when mandatory kwargs start to make sense

sergiusens (approved):
Looks good.

PR #89, opened 2022-06-21
Improved default global options: removed --trace and added --verbosity as a way to set all modes (CRAFT-1133).

facundobatista:


craft_cli/dispatcher.py
```diff
@@ -359,17 +359,24 @@ def pre_parse_args(
...
         elif global_args["verbose"]:
             emit.set_mode(EmitterMode.VERBOSE)
-        elif global_args["trace"]:
-            emit.set_mode(EmitterMode.TRACE)
+        elif global_args["verbosity"]:
+            try:
+                verbosity_level = EmitterMode[global_args["verbosity"].upper()]
+            except KeyError:
+                raise ArgumentParsingError(  # pylint: disable=raise-missing-from
+                    "Bad verbosity level; allowed are "
```
sergiusens, line 376:
```suggestion
                    "Bad verbosity level; valid values are "
```

docs/explanations.rst
```diff
@@ -25,7 +25,7 @@ How Craft CLI manage the application logs
 
 Unless overridden when ``emit`` is initiated (see :ref:`how to do that <howto_other_logfile>`), the application logs will be managed by the Craft CLI library, according to the following rules:
 
-- one log file is always produced for each application run (only exposed to the user if the application ends in error or a verbose run was requested by ``--verbose`` or ``--trace``), naming the files with a timestamp so they are unique
+- one log file is always produced for each application run (only exposed to the user if the application ends in error or a verbose run was requested by ``--verbose`` or ``--verbosity=trace``), naming the files with a timestamp so they are unique
```
sergiusens, line 28: what about --verbosity=debug ?

craft_cli/dispatcher.py
```diff
@@ -359,17 +359,24 @@ def pre_parse_args(
...
+                    "'quiet', 'brief', 'verbose', 'debug' and 'trace'."
+                )
+            emit.set_mode(verbosity_level)
```
sergiusens, line 379: is it correct that set_mode still does not understand  DEBUG and is that expected for this PR?

craft_cli/dispatcher.py
facundobatista: Improved, thanks!

docs/explanations.rst
facundobatista: I was considering it an example... I'll be more explicit and specify it as an example. I don't want to be 100% complete, otherwise I'll need to also mention --verbosity=verbose, and it will be a mess regarding readability.

craft_cli/dispatcher.py
facundobatista: `set_mode` (like `init`) will not crash, they will set `emit`'s level to DEBUG, but the Emitter will not behave yet as we want for this level, that's intended for the next branch.

sergiusens (approved):

PR #91, opened 2022-06-27
Improve the _StreamContextManager to show timestamps only when it has to (CRAFT-1152).

facundobatista:


cmatsuoka (approved):
Awesome, thanks!

PR #93, opened 2022-06-28
Transition middle important messages (CRAFT-1135).

facundobatista:
This is from `.message(..., intermediate=True)` to `.progress(...,
permanent=True)`.

Also documented changes from this (and previous!) PRs.

### DIFF
--- a/docs/explanations.rst
+++ b/docs/explanations.rst
@@ -101,11 +101,9 @@ Regular messages
 
 The ``message`` method is for the final output of the running command.
 
-If there is important information that needs to be shown to the user in the middle of the execution (and not overwritten by other messages) this method can be also used but passing ``intermediate=True``:
-
 ::
 
-    def message(self, text: str, intermediate: bool = False) -> None:
+    def message(self, text: str) -> None:
 
 E.g.::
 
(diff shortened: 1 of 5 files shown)


docs/explanations.rst
```diff
@@ -152,18 +152,49 @@ E.g.::
                     break
 
 
-Trace/debug messages
-~~~~~~~~~~~~~~~~~~~~
+Verbose messages
+~~~~~~~~~~~~~~~~
+
+Useful to provide more information to the user that shouldn't be exposed when in brief mode for clarity and simplicity.
```
cmatsuoka, line 158: Nitpick: in documentation, I would say: "Verbose messages are useful to..."

docs/explanations.rst
```diff
@@ -233,3 +264,186 @@ After that, is just a matter of running the file specifying which example to use
...
+     - BRIEF
+     - VERBOSE
+     - DEBUG
+     - TRACE
+     - also to logfile
+   * - ``.message(...)``
+     - --
+     - | stdout
+       | permanent
+       | plain
```
cmatsuoka, line 290: Some of the "plain" strings have different styling but I think the github diff is confused with markup instead of showing verbatim content. I'm assuming this is correct in the source code.

cmatsuoka (approved):
Nice! Added some notes but it's mostly nitpicking.

docs/explanations.rst
facundobatista: The table builds ok and consistent, yes. Thanks!

docs/explanations.rst
facundobatista: Improved! Thanks!

PR #99, opened 2022-06-30
Support the unlinking of files that were just removed in the logpath helper (CRAFT-1125).

### LINKED ISSUE #84: Log garbage collection is not safe for concurrent use
References https://github.com/goreleaser/goreleaser/issues/1715

We publish Snaps as part of our release CI flow. We do this using [`goreleaser`](https://goreleaser.com/), which will use parallel `snapcraft` instances in order to upload the Snaps. The problem is that this library (`craft-cli`) is not safe for concurrent use and this behavior carries forward to `snapcraft`.

The main issue that I've met so far in `craft-cli` is that the automatic log garbage collection is not safe for concurrent use:
https://github.com/canonical/craft-cli/blob/76e36c0648a45f018ef48e51fd696ab126044f7d/craft_cli/messages.py#L125-L129

If you are uploading 5 Snaps in parallel, you will have 5 logs being generated in parallel. As they finish the upload all 5 of them will list the contents of the logs directory and attempt to delete them, but since you cannot delete the same file twice, one of the commands will fail.

If requiring Python 3.8 or higher is ok, using the `missing_ok` parameter of `unlink` probably will suffice.

facundobatista:
Fixes #84.

### DIFF
--- a/tests/unit/test_messages_helpers.py
+++ b/tests/unit/test_messages_helpers.py
@@ -146,7 +146,7 @@ def test_getlogpath_supports_missing_file_to_unlink(test_log_dir, monkeypatch):
 
     def mitm(self, *a, **k):
         if self.name.startswith("testapp") and self.name.endswith(".log"):
-            # it's trying to remove the log file, let's remove it us first
+            # it's trying to remove the log file, let's remove it first
             orig_method(self)
         orig_method(self, *a, **k)
 
--- a/tests/unit/test_messages_helpers.py
+++ b/tests/unit/test_messages_helpers.py
@@ -18,6 +18,7 @@
 
 import datetime
 import logging
+import pathlib
 import re
 import sys
 import time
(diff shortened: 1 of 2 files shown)


tests/unit/test_messages_helpers.py
```diff
@@ -130,6 +131,31 @@ def test_getlogpath_exceeds_rotation_limit(test_log_dir, monkeypatch):
...
+        fpath.touch()
+        previous_fpaths.append(fpath)
+        time.sleep(0.01)  # sleep a little so different log files have different timestamps
+
+    # hook a MITM function to remove the file before it was unlinked
+    orig_method = pathlib.Path.unlink
+
+    def mitm(self, *a, **k):
+        if self.name.startswith("testapp") and self.name.endswith(".log"):
+            # it's trying to remove the log file, let's remove it us first
```
sergiusens, line 149: the "let's" contraction is "let us"
```suggestion
            # it's trying to remove the log file, let's remove it first
```

tests/unit/test_messages_helpers.py
```diff
@@ -130,6 +131,31 @@ def test_getlogpath_exceeds_rotation_limit(test_log_dir, monkeypatch):
     assert present_logs == previous_fpaths[1:] + [new_fpath]
 
 
+def test_getlogpath_supports_missing_file_to_unlink(test_log_dir, monkeypatch):
+    """It's ok if the file to unlink was previously removed."""
```
sergiusens, line 135: is this not more of a code comment? These docstrings are sometimes used as the summary of a test and this does not look like that.

facundobatista: I use tests docstrings to explain what the test does (or try to).

Our infrastructure (pytest) shows `./E/F` in brief mode or the test function name when verbose.

tests/unit/test_messages_helpers.py
facundobatista: Yes! thanks

tests/unit/test_messages_helpers.py
sergiusens: ok, this however does not sound like an explanation but more like a clarification

sergiusens (approved):

PR #100, opened 2022-07-04
Behave as 'captured output' when the (fake) terminal has zero columns (CRAFT-1128).

facundobatista:
The fix can be tested IRL by using `expect` with the following script:
```
set stty_init "rows 100 cols 0"
spawn env/bin/python examples.py 1
expect "The meaning of life is 42."
```


cmatsuoka (approved):
Simple and reasonable fix.

Issue #102, opened 2022-07-06
"ValueError: filedescriptor out of range in select()" error message when building big snaps


sergio-costas:
While building from scratch the 'add_adwaita' pull request of 'gnome-sdk' repository (https://github.com/ubuntu/gnome-sdk/pull/37), during the "prime" stage, a lot of these messages are shown:

```
2022-07-06 10:18:08.492 Executing parts lifecycle: prime libadwaita                                                                                                                                       
2022-07-06 10:18:08.492 Executing action                                                                                                                                                                  
Exception in thread Thread-512:
2022-07-06 10:18:08.493 execute action libadwaita:Action(part_name='libadwaita', step=Step.PRIME, action_type=ActionType.RUN, reason=None, project_vars=None)                                             Traceback (most recent call last):
  File "/snap/snapcraft/7855/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/snap/snapcraft/7855/lib/python3.8/site-packages/craft_cli/messages.py", line 541, in run
    self._run_posix()
  File "/snap/snapcraft/7855/lib/python3.8/site-packages/craft_cli/messages.py", line 510, in _run_posix
    rlist, _, _ = select.select([self.read_pipe], [], [], 0.1)
ValueError: filedescriptor out of range in select()

2022-07-06 10:18:08.584 Executed: prime libadwaita                                                                                                                                                        
2022-07-06 10:18:08.584 Executing parts lifecycle: prime p11-kit                                                                                                                                          
2022-07-06 10:18:08.584 Executing action                                                                                                                                                                  
Exception in thread Thread-513:
2022-07-06 10:18:08.584 execute action p11-kit:Action(part_name='p11-kit', step=Step.PRIME, action_type=ActionType.RUN, reason=None, project_vars=None)                                                   Traceback (most recent call last):
  File "/snap/snapcraft/7855/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/snap/snapcraft/7855/lib/python3.8/site-packages/craft_cli/messages.py", line 541, in run
    self._run_posix()
  File "/snap/snapcraft/7855/lib/python3.8/site-packages/craft_cli/messages.py", line 510, in _run_posix
    rlist, _, _ = select.select([self.read_pipe], [], [], 0.1)
ValueError: filedescriptor out of range in select()

2022-07-06 10:18:08.630 Executed: prime p11-kit                                                                                                                                                           
[...]

PR #103, opened 2022-07-06
Close pipes when the thread ends

sergio-costas:
The _PipeReaderThread() class creates two pipes to manage input/output, and uses SELECT() to know when there is data to be read. Unfortunately, those two pipes aren't closed when the thread ends, which means that the FDs of the pipes of new classes will grow up and up. The problem is that SELECT has a limit in the FD value that can monitor, usually 1024, which means that any FD with a value greater than that can't be monitored by it, and will show the error if tried to:

`  ValueError: filedescriptor out of range in select()`

With small snaps this is not a problem because they don't create so many objects of this kind as to reach the limit, but with very big snaps (like Gnome-42-2204-SDK, with more than 50 binaries), it can be achieved.

This MR partially fixes this by closing the FDs when the thread ends. Anyway, the best solution is to also use POLL instead of SELECT, because it doesn't have that limit (this is implemented in https://github.com/canonical/craft-cli/pull/104 , which should be merged along this one).

fix https://github.com/canonical/craft-cli/issues/102

- [X] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?

-----

### DIFF
--- a/tests/unit/test_messages_stream_cm.py
+++ b/tests/unit/test_messages_stream_cm.py
@@ -204,7 +204,7 @@ def test_pipereader_chunk_assembler(recording_printer, monkeypatch):
 
 
 def test_no_fail_if_big_number_is_used(recording_printer):
-    """Ensures that all the resources are freed on exit."""
+    """Ensures that opening and closing a big number of objects doesn't fail."""
     # The limit is 1024 both in Linux and BSD, but each object opens two FDs
     for counter in range(514):
         print(counter, end="\r")  # just to use the variable
@@ -213,3 +213,14 @@ def test_no_fail_if_big_number_is_used(recording_printer):
         prt.stop()
 
     assert True
+
+
+def test_ensure_pipes_are_closed(recording_printer):
+    """Ensures that all the resources are freed on exit."""
+    prt = _PipeReaderThread(recording_printer, sys.stdout, use_timestamp=False)
+    prt.start()
+    prt.stop()
+    with pytest.raises(Exception):
+        os.fstat(prt.read_pipe)
+    with pytest.raises(Exception):
+        os.fstat(prt.write_pipe)
--- a/tests/unit/test_messages_stream_cm.py
+++ b/tests/unit/test_messages_stream_cm.py
@@ -201,3 +201,15 @@ def test_pipereader_chunk_assembler(recording_printer, monkeypatch):
     msg1, msg2 = recording_printer.written_terminal_lines
     assert msg1.text == ":: ------abcde---"
     assert msg2.text == ":: otherline---"
+
+
+def test_no_fail_if_big_number_is_used(recording_printer):
+    """Ensures that all the resources are freed on exit."""
+    # The limit is 1024 both in Linux and BSD, but each object opens two FDs
+    for counter in range(514):
+        print(counter, end="\r")  # just to use the variable
+        prt = _PipeReaderThread(recording_printer, sys.stdout, use_timestamp=False)
+        prt.start()
+        prt.stop()
+
+    assert True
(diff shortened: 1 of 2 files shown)


sergio-costas:
Tested it by "backporting" it to craft-cli 0.6.0, creating snapcraft with the patched version, and built gnome-42-2204-sdk ("add_adwaita" MR)

sergio-costas:
I'm also preparing a patch that uses poll() instead of select.

facundobatista (changes requested):
Hello! thanks for this work! we would need a test for this to land, though.

Thanks!

sergio-costas:
I'm trying to execute the tests (even from MAIN branch) and I always receive these errors:

imagen

The odd thing is that if I get the values of `EmitterMode` inside the tests, they are: [<EmitterMode.QUIET: 1>, <EmitterMode.NORMAL: 2>, <EmitterMode.VERBOSE: 3>, <EmitterMode.TRACE: 4>]

I tried to add some debug both in `craft_cli/__init__.py` and `craft_cli/messages.py` to see what values are put there, but those files aren't run when called the tests (I tried both with `make test-units` and `pytest-3 tests/unit`, and got the same results).

What am I doing wrong?

sergio-costas:
Ok, I found the problem: I had a global version installed in my system.

sergio-costas:
Mmm... I'm still receiving a lot of error messages when executing the tests... they say that "emit" hasn't been initialized...

sergio-costas:
> Hello! thanks for this work! we would need a test for this to land, though.

Added a test. It fails with the code in MAIN, but does pass with the patch.

tests/unit/test_messages_stream_cm.py
```diff
@@ -201,3 +201,15 @@ def test_pipereader_chunk_assembler(recording_printer, monkeypatch):
     msg1, msg2 = recording_printer.written_terminal_lines
     assert msg1.text == ":: ------abcde---"
     assert msg2.text == ":: otherline---"
+
+
+def test_no_fail_if_big_number_is_used(recording_printer):
```
facundobatista, line 206: Hello! Thanks for the test. 

I'd suggest a more "unittesty" one: just start the _PipeReaderThread, stop it, and then assert that both pipes are closed. (the current approach has the problem that is testing the side-effect, which may change if we modify internals... for example, if we start using one pipe and forgot to close it, this will still pass...)

Thanks again!

facundobatista (changes requested):
Thanks for test! There I suggested a simplification of the test.

tests/unit/test_messages_stream_cm.py
sergio-costas: Should I add another unitary test, or replace the current one?

I think that it would be better to do both...

sergio-costas: Added the new test.

facundobatista (changes requested):

tests/unit/test_messages_stream_cm.py
```diff
@@ -201,3 +201,26 @@ def test_pipereader_chunk_assembler(recording_printer, monkeypatch):
...
+
+
+def test_ensure_pipes_are_closed(recording_printer):
+    """Ensures that all the resources are freed on exit."""
+    prt = _PipeReaderThread(recording_printer, sys.stdout, use_timestamp=False)
+    prt.start()
+    prt.stop()
+    with pytest.raises(Exception):
+        os.fstat(prt.read_pipe)
+    with pytest.raises(Exception):
```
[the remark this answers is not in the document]
sergio-costas: Done!

facundobatista (approved):
Wonderful, thanks!!

Issue #105, opened 2022-07-07
Unitary tests fail when run from command line


sergio-costas:
Running `make test-units` from the command line, or `pytest-3 tests/unit` return several test failures, all with the error:
```
    RuntimeError: Emitter needs to be initiated first
```

facundobatista:
If we have a race condition or something similar, we need to understand what.

Let's try to isolate it

Let's grab one test that failed in that big output and see what happens... what's your output for the following?

```
$ pytest -svv tests/unit/test_dispatcher.py -k test_dispatcher_pre_parsing
================================================================================ test session starts =================================================================================
platform linux -- Python 3.9.7, pytest-7.1.2, pluggy-1.0.0 -- /home/facundo/canonical/craft-cli/env/bin/python3
cachedir: .pytest_cache
rootdir: /home/facundo/canonical/craft-cli, configfile: setup.cfg
plugins: mock-3.7.0, xdist-2.5.0, subprocess-1.4.1, forked-1.4.0, craft-cli-1.0.0
collected 86 items / 85 deselected / 1 selected

tests/unit/test_dispatcher.py::test_dispatcher_pre_parsing PASSED

========================================================================== 1 passed, 85 deselected in 0.03s ==========================================================================
```

sergio-costas:
This is my output:

```
$ pytest -svv tests/unit/test_dispatcher.py -k test_dispatcher_pre_parsing
=========================================================================================== test session starts ===========================================================================================
platform linux -- Python 3.10.4, pytest-6.2.5, py-1.10.0, pluggy-0.13.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/raster/workspace/craft-cli, configfile: setup.cfg
plugins: cov-3.0.0, requests-mock-1.9.3, pylint-0.18.0, mock-3.6.1
collected 86 items / 85 deselected / 1 selected                                                                                                                                                           

tests/unit/test_dispatcher.py::test_dispatcher_pre_parsing FAILED

================================================================================================ FAILURES =================================================================================================

```

sergio-costas:
Ok, my fault... I forgot to enable the venv. Closing.

Sorry.

PR #106, opened 2022-07-07
Fix unitary tests in local

sergio-costas:
When running "make test-units" from a shell, several tests fail
with the error:

`  RuntimeError: Emitter needs to be initiated first`

This patch fixes this by initializing the 'emit' object.

Fix https://github.com/canonical/craft-cli/issues/105

-----

### DIFF
--- a/tests/unit/test_dispatcher.py
+++ b/tests/unit/test_dispatcher.py
@@ -31,6 +31,8 @@ from tests.factory import create_command
 
 # --- Tests for the Dispatcher
 
+emit.init(EmitterMode.DEBUG, "testappname", "greeter")
+
 
 def test_dispatcher_help_init():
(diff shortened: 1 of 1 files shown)


facundobatista:
Which test is failing for you?

facundobatista (changes requested):
Hello! 

I don't like initiating the emitter at module level "just in case", if we have a race condition or something similar, we need to understand what is going on.

FTR, local tests run ok for me and other people. Let's follow up on this in the issue or in the chat. Thanks!

sergio-costas:
> Which test is failing for you?

This is the output: https://pastebin.canonical.com/p/YmwnkSF7Hq/

sergio-costas:
My fault: I forgot to enable the venv.

Sorry.

PR #109, opened 2022-07-20
Better tested help parsing function, with a bug fixed.

facundobatista:
In preparaion to the "help with markdown output" work I better tested the help parsing function, and fixed a bug where the error was not properly raises when `--all` was used with some extra info.


mr-cal (approved):
Looks good!

PR #110, opened 2022-07-20
Helper function to process commands overviews and leave markdown text.

facundobatista:
This is an isolated helper still not used, in preparation for the next
branch of "help in markdown format".

### DIFF
--- a/craft_cli/helptexts.py
+++ b/craft_cli/helptexts.py
@@ -70,6 +70,42 @@ def _build_item(title: str, text: str, title_space: int) -> List[str]:
     return result
 
 
+def process_overview_for_markdown(text: str) -> str:
+    """Process a regular overview to be rendered with markdown.
+
+    In detail:
+
+    - Join all lines for the same paragraph (as wrapping is responsibility of the renderer)
+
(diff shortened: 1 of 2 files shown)


craft_cli/helptexts.py
```diff
@@ -70,6 +70,42 @@ def _build_item(title: str, text: str, title_space: int) -> List[str]:
...
+    In detail:
+
+    - Join all lines for the same paragraph (as wrapping is responsibility of the renderer)
+
+    - Dedent and wrap with triple-backtick all indented blocks
+
+    Paragraphs are separated by empty lines
+    """
+    lines = [x.rstrip() for x in text.strip().split("\n")]
+    blocks: List[List[str]] = [[]]
```
mr-cal, line 85: It would be nice to have a comment defining what a `block` is here or in the docstring.  (Is it a paragraph?  And is a paragraph just a set of lines not separated by an empty line?)

mr-cal (approved):
Looks good!

craft_cli/helptexts.py
facundobatista: Sure, added!

PR #111, opened 2022-07-21
Added the option of showing the help for a command in markdown format (CRAFT-1207).

facundobatista:


cmatsuoka (approved):
We may need additional processing to escape text containing markdown syntax or converting potentially problematic characters such as < or >, but the general implementation looks good. Specific problems can be addressed as needed.

PR #113, opened 2022-08-04
In the "command help" in markdown format, the Options items are built in a table so they are vertically aligned.

facundobatista:


craft_cli/helptexts.py
```diff
@@ -357,9 +347,14 @@ def _build_markdown_command_help(self, usage, overview, options, other_command_n
...
 
-        option_lines = ["## Options:"]
+        option_lines = [
+            "## Options:",
+            "| | |",
+            "|-|-|",
+        ]
         for title, text in options:
-            option_lines.extend(_build_item_markdown(title, text))
+            option_lines.append(f"| `{title}` | {text} |")
```
sergiusens, line 356: maybe a future change could be to make this a 3 column line so the long options and short ones are aligned and would be easy to spot which one has which, I understand this is a bigger change in the help parser though

sergiusens (approved):

sergiusens:
looks gooood

craft_cli/helptexts.py
facundobatista: Yes! I opened #114 to track it.

PR #126, opened 2022-10-07
printer: enhance ephemeral message handover after stop() or pause()

flotter:
It is a very common pattern for craft tools to start the session
directly on a host machine, but launch another craft instance to
do the actual work inside LXD or Multipass.

In order for a seamless CLI experience (which looks identical
between a 'destructive build' on the host, LXD or Multipass),
we need support for progress messages (ephemeral=True) to seamlessly
cross the boundaries between the host and the container instance of
of the craft tool.

Please see examples.py #26 for an example.

Add support for _Printer to respect ephemeral messages when
stopping. On printer stop(), these messages are no longer terminated
with a newline, but instead the printer will clear and reset the
cursor to the start of the line.

This behaviour change will allow subsequent terminal writes by a
different process the ability to reuse a terminal line previously
used as ephemeral output.

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -234,7 +234,7 @@ class _Printer:
         # open the log file (will be closed explicitly later)
         self.log = open(log_filepath, "at", encoding="utf8")  # pylint: disable=consider-using-with
 
-        # keep account of output streams with unfinished lines
+        # keep account of output terminal streams with unfinished lines
         self.unfinished_stream: Optional[TextIO] = None
 
         # run the spinner supervisor
@@ -439,7 +439,18 @@ class _Printer:
         if not TESTMODE:
             self.spinner.stop()
         if self.unfinished_stream is not None:
-            print(flush=True, file=self.unfinished_stream)
+            # With unfinished_stream set, the prv_msg object is valid.
+            if self.prv_msg.ephemeral:  # type: ignore
+                # If the last printed message is of 'ephemeral' type, the stop
+                # request must clean and reset the line.
+                cleaner = " " * (_get_terminal_width() - 1)
+                line = "\r" + cleaner + "\r"
+                print(line, end="", flush=True, file=self.prv_msg.stream)  # type: ignore
+            else:
+                # The last printed message is permanent. Leave the cursor on
+                # the next clean line.
+                print(flush=True, file=self.unfinished_stream)
+
         self.log.close()
         self.stopped = True
 
--- a/tests/unit/test_messages_integration.py
+++ b/tests/unit/test_messages_integration.py
@@ -378,6 +378,46 @@ def test_progressbar_brief_terminal(capsys, monkeypatch):
     assert_outputs(capsys, emit, expected_err=expected_screen, expected_log=expected_log)
 
 
+@pytest.mark.parametrize("output_is_terminal", [True])
+def test_progressbar_brief_permanent_terminal(capsys, monkeypatch):
+    """Show a progress bar in brief mode."""
+    # fake size so lines to compare are static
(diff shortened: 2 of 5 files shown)


craft_cli/messages.py
```diff
@@ -439,7 +439,18 @@ def stop(self) -> None:
         if not TESTMODE:
             self.spinner.stop()
         if self.unfinished_stream is not None:
-            print(flush=True, file=self.unfinished_stream)
+            # With unfinished_stream set, the prv_msg object is valid.
+            if self.prv_msg.ephemeral:  # type: ignore
+                # If the last printed message is of 'ephemeral' type, the stop
+                # request must clean and reset the line.
+                cleaner = " " * (_get_terminal_width() - 1)
```
facundobatista, line 446: I think that the `- 1` does not correspond here, because you're not actually showing the cursor at the end of the terminal (in its last column), because of the second CR.

tests/unit/test_messages_integration.py
```diff
@@ -230,11 +230,18 @@ def test_progress_brief_terminal(capsys):
...
     emit.ended_ok()
 
-    expected = [
+    expected_term = [
         Line("The meaning of life is 42.", permanent=False),
-        Line("Another message.", permanent=True),  # stays as it's the last message
+        Line("Another message.", permanent=False),
+        # This cleaner line is inserted by the printer stop
+        # sequence to reset the last ephemeral print to terminal.
+        Line("", permanent=False),
```
facundobatista, line 238: Please duplicate this test and have one as before (with permanent last line) and one as you're suggesting (with ephimeral last line). This way we ensure that this will work correctly for ever in BOTH situations.

tests/unit/test_messages_integration.py
```diff
@@ -358,7 +365,10 @@ def test_progressbar_brief_terminal(capsys, monkeypatch):
         Line("Uploading stuff [████████████████████████       ] 1400/1788", permanent=False),
         Line("Uploading stuff [███████████████████████████████] 1788/1788", permanent=False),
         Line("Uploading stuff (<---)", permanent=False),
-        Line("And so on", permanent=True),
+        Line("And so on", permanent=False),
```
facundobatista, line 368: Duplicate this one too, please, as suggested above.

facundobatista (changes requested):
Awesome, thanks for this! Annotated a couple of details for improvement. Thanks again!

craft_cli/messages.py
flotter: True.

Just note that we are technically "cleaning" one character more than what would ever be written by an ephemeral message. If you are happy with the inconsistency, then we can clean the entire line here.

facundobatista: yes, it's fine, because the ephemeral message actually uses one more position that our cleanup here (the cursor)

tests/unit/test_messages_integration.py
flotter: I believe the test ```test_progress_brief_permanent``` covers what you are asking for. Let me know if I made a mistake.

tests/unit/test_messages_integration.py
flotter: Created a ```permanent=True``` version to test the case I previously removed.

tests/unit/test_messages_integration.py
facundobatista: No, the other one is about the first (or intermediate) message. But in its way it's testing that, so we're ok.

facundobatista (approved):
Wonderful, thanks for this work!!!

facundobatista:
@flotter note that some tests are failing...

PR #131, opened 2023-01-10
(messages) Show messages from subprocess in an ephemeral way when in brief mode. (CRAFT-1542)

facundobatista:
Also added an example to show this (and improved the whole examples machinery to accept parameters, to make it easier to use).

Added a couple of small changes because pyright/pylint, and will split the big `messages.py` in a following branch (without semantic changes in it).

Documentation updated.

No change needed from the applications using the library.

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -652,7 +652,7 @@ class _StreamContextManager:
         printer_flags = {
             "use_timestamp": use_timestamp,
             "ephemeral": ephemeral_mode,
-            "end_line": not ephemeral_mode
+            "end_line": not ephemeral_mode,
         }
 
         # show the intended text (explicitly asking for a complete line) before passing the
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -649,11 +649,11 @@ class _StreamContextManager:
     ):
         # prepare the printer flags for the initial message and everything produced
         # by the pipe reader
-        printer_flags = {"use_timestamp": use_timestamp}
-        if ephemeral_mode:
-            printer_flags.update(ephemeral=True, end_line=False)
-        else:
-            printer_flags.update(ephemeral=False, end_line=True)
+        printer_flags = {
+            "use_timestamp": use_timestamp,
+            "ephemeral": ephemeral_mode,
+            "end_line": not ephemeral_mode
+        }
 
         # show the intended text (explicitly asking for a complete line) before passing the
         # output command to the pip-reading thread
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -1,5 +1,7 @@
+# XXX Facundo 2022-01-10: will split this module in a subsequent branch
+# pylint: disable=too-many-lines
 #
-# Copyright 2021-2022 Canonical Ltd.
+# Copyright 2021-2023 Canonical Ltd.
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -39,7 +41,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import lru_cache
-from typing import Literal, Optional, TextIO, Union
+from typing import Dict, Literal, Optional, TextIO, Union
 
 import platformdirs
 
@@ -526,9 +528,11 @@ class _PipeReaderThread(threading.Thread):
     # byte used to unblock the reading (under Windows)
     UNBLOCK_BYTE = b"\x00"
 
-    def __init__(self, printer: _Printer, stream: Optional[TextIO], use_timestamp: bool):
+    def __init__(
+        self, printer: _Printer, stream: Optional[TextIO], printer_flags: Dict[str, bool]
+    ):
         super().__init__()
-        self.use_timestamp = use_timestamp
+        self.printer_flags = printer_flags
 
         # prepare the pipe pair: the one to read (used in the thread core loop) and the
         # one which is to be written externally (and also used internally under windows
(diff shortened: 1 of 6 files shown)


tests/unit/test_messages_integration.py
```diff
@@ -719,6 +712,70 @@ def test_third_party_output_quietish_modes(capsys, tmp_path, mode):
...
+    """Manage the streams produced for sub-executions, brief mode, to the terminal."""
+    # something to execute
+    script = tmp_path / "script.py"
+    script.write_text(
+        textwrap.dedent(
+            """
+        import sys
+        print("foobar out", flush=True)
+        print("foobar err", file=sys.stderr, flush=True)
+    """
```
tigarmo, line 726: weird indentation of quotes

craft_cli/messages.py
```diff
@@ -526,9 +528,11 @@ class _PipeReaderThread(threading.Thread):
     # byte used to unblock the reading (under Windows)
     UNBLOCK_BYTE = b"\x00"
 
-    def __init__(self, printer: _Printer, stream: Optional[TextIO], use_timestamp: bool):
+    def __init__(
+        self, printer: _Printer, stream: Optional[TextIO], printer_flags: Dict[str, bool]
```
tigarmo, line 532: it's just a suggestion but have you considered using a dataclass instead of a `Dict[str,bool]` for the printer flags? It's more self-documenting and protects against "wrong" dict keys

tigarmo (approved):
LGTM

craft_cli/messages.py
facundobatista: Yes, but it would mean that everytime a message is shown/produced (for every line in the subprocess), the dict will have to be built and rebuilt to call the printer .show() method... just a dict sounded more natural for this.

tests/unit/test_messages_integration.py
facundobatista: Fixed!

craft_cli/messages.py
```diff
@@ -633,15 +639,28 @@ def stop(self) -> None:
...
+        use_timestamp: bool,
+        ephemeral_mode: bool,
     ):
+        # prepare the printer flags for the initial message and everything produced
+        # by the pipe reader
+        printer_flags = {"use_timestamp": use_timestamp}
+        if ephemeral_mode:
+            printer_flags.update(ephemeral=True, end_line=False)
+        else:
+            printer_flags.update(ephemeral=False, end_line=True)
```
lengau, line 656:
```suggestion
        printer_flags = {
            "use_timestamp": use_timestamp,
            "ephemeral": ephemeral_mode,
            "end_line": not ephemeral_mode
        }
```

craft_cli/messages.py
lengau: A possibility would be to take in the dataclass and convert it to a dict in `__init__`, but I'm not sure it's worth the hassle

lengau (approved):
This looks good to me. If we start to have weird issues with the pipes on Windows we can try re-adding (and perhaps shortening) the sleep time and then document that it's a workaround for whatever those issues are.

tests/unit/test_messages_integration.py
```diff
@@ -1168,18 +1225,22 @@ def test_logging_after_closing(capsys, logger):
...
     tstamp += int(msec) / 1000
     return tstamp
 
 
-@pytest.mark.parametrize("loops, sleep, max_repetitions", [
-    (100, .01, 30),
-    (1000, .001, 100),
-    (10, .1, 2),
-])
+@pytest.mark.skipif(sys.platform == "win32", reason="Temporary skip, need to investigate why slow")
```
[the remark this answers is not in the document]
facundobatista: agh

tests/unit/test_messages_stream_cm.py
```diff
@@ -112,10 +124,42 @@ def test_streamcm_init_with_stream_and_timestamp(recording_printer, stream):
...
+
+    # initial message
+    (msg,) = recording_printer.written_terminal_lines  # pylint: disable=unbalanced-tuple-unpacking
+    assert msg.stream == stream
+    assert msg.text == "initial text"
+    assert msg.use_timestamp is False
+    assert msg.end_line is False
+    assert msg.ephemeral is True
+    assert msg.bar_progress is None
+    assert msg.bar_total is None
```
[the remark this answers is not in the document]
facundobatista: Nice! Didn't know about pytest-check; in this case it does not worth it, as these checks are more "sanity" ones...

tests/unit/test_messages_integration.py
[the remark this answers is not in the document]
facundobatista: removed the skip, tests still green...

craft_cli/messages.py
facundobatista: I like it!!

craft_cli/messages.py
```diff
@@ -633,15 +639,28 @@ def stop(self) -> None:
...
+        stream: Optional[TextIO],
+        use_timestamp: bool,
+        ephemeral_mode: bool,
     ):
+        # prepare the printer flags for the initial message and everything produced
+        # by the pipe reader
+        printer_flags = {
+            "use_timestamp": use_timestamp,
+            "ephemeral": ephemeral_mode,
+            "end_line": not ephemeral_mode
```
lengau, line 655: Oops, that linting error was my fault. Sorry!
```suggestion
            "end_line": not ephemeral_mode,
```

lengau:
The test failure is probably a continuation of [this pypi incident](https://status.python.org/incidents/j369h2rkqsgz) from earlier this week.

craft_cli/messages.py
facundobatista: No worries, fixed.

craft_cli/messages.py
facundobatista: Right.

PR #137, opened 2023-02-12
packaging: replace deprecated setuptools directive

jugmac00:
See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-license-file

Committed via https://github.com/asottile/all-repos


jugmac00:
The linter errors are unrelated.

Issue #139, opened 2023-02-28
emit.pause is creating new Printer, but not updating printer for logging


rgildein:
The `emit.pause` is stopping printer and than creating new one, however the `emit._log_handler` is still using old printer.

How to reproduced.
```python
import logging

from craft_cli.messages import EmitterMode, emit

logger = logging.getLogger(__name__)


def main():
    emit.init(EmitterMode.VERBOSE, "test", "test emit.pause")
    logger = logging.getLogger()
    logger.setLevel(logging.DEBUG)

    # test part
    emit.message("EMIT: start")
    logger.info("logging: start")
    with emit.pause():
        print("pause")

    logger.info("logging: stop")
    emit.message("EMIT: stop")


if __name__ == "__main__":
    main()
```

### actual behavior
```bash
(.venv) x1:➜  craft-cli git:(main) ✗ python3 -m test
test emit.pause                                                                                                                                                                      
Logging execution to '/home/rgildein/.local/state/test/log/test-20230228-182816.590736.log'                                                                                          
EMIT: start                                                                                                                                                                          
logging: start                                                                                                                                                                       
pause
EMIT: stop                                                                                                                                                                     [0.17s]
```
### expected bahvior
```bash
(.venv) x1:➜  craft-cli git:(main) ✗ python3 -m test
test emit.pause                                                                                                                                                                      
Logging execution to '/home/rgildein/.local/state/test/log/test-20230228-182828.684952.log'                                                                                          
EMIT: start                                                                                                                                                                          
logging: start                                                                                                                                                                       
pause
logging: stop                                                                                                                                                                        
EMIT: stop           
```
Issue is in this [line](https://github.com/canonical/craft-cli/blob/main/craft_cli/messages.py#L618) of code where new printer is created, but it's not updated for `self._log_handler.printer`.

PR #140, opened 2023-02-28
fix new printer for logging handler

rgildein:
I'm not sure if this is the best solution here or we should use something like `printer.start` / `printer.recreate`.
I also think some unit tests would be nice and will add them if this solution is approved.

fixes: #139

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -615,7 +615,7 @@ class Emitter:
             yield
         finally:
             self._stopped = False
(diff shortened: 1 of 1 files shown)


lengau:
Closing and reopening to re-run tests and linting

lengau (approved):
Thanks for this fix!

The linting issue is fixed in #142

lengau:
Closing/reopening for check changes

PR #145, opened 2023-06-06
Merge starbase/main

tigarmo:
This PR integrates craft-cli with the current "state of the art" implemented by Starbase. It's a huge changeset because it's a merge of two distinct repositories, but the idea is that future merges will be easier/less painful.

The general strategy was this: First I merged `starbase/main` into this branch, and for the conflicts I preferred starbase's version for most cases, unless the fix was very straightforward. This merge is commit 13c852fd4e5f6ec77ed145b6b253d38f16e5d769.

After that, I added commits to migrate craft-cli's way of doing things to starbase's. For example, the first commit after the merge removes `setup.py` and `setup.cfg` and adds the config to `pyproject.toml`. This migration was done manually and I inspected/tested the resulting wheel, etc. The next commits add fixes for the individual linters, coverage, and a bigger one for the docs. Then I removed the files that are no longer needed (like the Makefiles).

Hopefully, the task of reviewing per-commit starting at that merge (or the one right after it, since it's hard to review the merge) won't be too onerous because each commit is reasonably self-contained; if it is, tell me and I can see about opening multiple PRs.


craft_cli/dispatcher.py
```diff
@@ -334,7 +334,9 @@ def _build_no_command_error(self, missing_command: str) -> str:
         msg = f"no such command {missing_command!r}{extra_similar}"
         return self._help_builder.get_usage_message(msg)
 
-    def _parse_options(self, defined_arguments, sysargs):  # pylint: disable=too-many-branches
+    def _parse_options(
```
lengau, line 337: We might want to take the type checker related changes and put them in their own PR. (As long as CI still succeeds)

HACKING.rst
```diff
@@ -0,0 +1,80 @@
+*********
```
mr-cal: I see this gets deleted in a subsequent commit.  Do you think it could be adapted for craft-cli?  I think it contains a lot of useful info about the tooling that isn't in the original README, which is minimal and outdated (like using `Make`)

mr-cal:
I added the `do-not-squash` label from [craft-parts](https://github.com/canonical/craft-parts/issues?q=label%3Ado-not-squash+)

mr-cal (commented):
Looks good!

README.rst
```diff
@@ -0,0 +1,33 @@
+********
```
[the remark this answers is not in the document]
lengau: I like it! https://github.com/canonical/starbase/issues/93

HACKING.rst
tigarmo: You're right! I thought about this a bit and I think that a good 'compromise' might be to finish this merge and then gave a separate issue/PR to update craft-cli's README and HACKING to integrate the content from starbase, because as you point out it's quite useful. What do you think?

pyproject.toml
```diff
@@ -95,6 +95,7 @@ exclude = [
 
 [tool.black]
 target-version = ["py38"]
+line-length = 99
```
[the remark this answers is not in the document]
tigarmo: #149 and #150

HACKING.rst
mr-cal: Yes, a separate issue/PR is fine with me!

sergiusens:
I removed branch protection rules to allow this to land when ready

sergiusens:
I have also disabled automerge to avoid any accidents

craft_cli/dispatcher.py
lengau: Hmmm yeah good point, there's no great way to do this. Retracted, let's keep it this way since it's less work and not particularly unclear.

lengau (approved):
I'm happy with this if @mr-cal is :-)

HACKING.rst
tigarmo: Done, #152

tigarmo:
@lengau @mr-cal I added a new commit re-adding pylint linting (in tox form), because without ruff the codebase is missing a "stylistic" linter. We can remove it again once we add ruff

PR #154, opened 2023-06-15
messages: support a 'prefix' in open_stream()

tigarmo:
If provided, this string replaces the default ":: " prefix that indicates that a message came from a subprocess. This can be used to provide extra context, indicate the step that is generating the info, etc.


lengau (approved):
Prepare to see :rocket: prefixes!

craft_cli/messages.py
```diff
@@ -579,8 +590,14 @@ def progress_bar(self, text: str, total: Union[int, float], delta: bool = True)
         return _Progresser(self._printer, total, text, stream, delta, use_timestamp, ephemeral)  # type: ignore
 
     @_active_guard()
-    def open_stream(self, text: str):
-        """Open a stream context manager to get messages from subprocesses."""
+    def open_stream(self, text: str, prefix: Optional[str] = None):
```
tigarmo, line 593: this is a good time to discuss whether we should add this extra `prefix` parameter, or use `text` as the prefix always.

lengau: Don't they do slightly different things? `text` provides an initial line of text, but `prefix` prefixes each subsequent line?

tigarmo: Yes, but because the initial line of text typically gets overwritten quickly, one could argue that it makes sense to unify them. For instance, if `text="Executing: build mypart"` then we could e.g. show `Executing: build mypart` first and then prefix each subsequent message with `Executing: build mypart :: `

PR #155, opened 2023-06-15
tools: call 'coverage' directly

tigarmo:
The problem: when using pytest-cov to integrate coverage with the test
run, there is a 'race' between pytest-cov and the pytest plugin provided
by craft-cli (the emitter). When craft-cli's plugin is loaded first, it
necessitates importing the main 'craft_cli' package, which in turn
imports dispatcher.py, messages.py, etc.
    
As a result, all the top-level lines (class and function definitions,
global-variables, etc) are loaded _before_ coverage starts running,
meaning that they are marked as 'hits' because they are never "loaded
again" during the tests. Instead, use coverage specifically so that
it's always set-up before the whole importing starts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,8 @@ xfail_strict = true
 
 [tool.coverage.run]
 branch = true
-parallel = true
+# Disable parallel coverage to preserve the data-file name so that 'coverage xml' finds it
+parallel = false
 omit = ["tests/**"]
 
 [tool.coverage.report]
--- a/tox.ini
+++ b/tox.ini
@@ -37,7 +37,12 @@ base = testenv, test
 description = Run unit tests with pytest
 labels =
     py38, py310, py311: tests, unit-tests
-commands = pytest {tty:--color=yes} --cov --cov-report=xml:results/coverage-{env_name}.xml --junit-xml=results/test-results-{env_name}.xml tests/unit {posargs}
+commands =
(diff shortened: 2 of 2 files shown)


tigarmo:
> Merging [#155](https://app.codecov.io/gh/canonical/craft-cli/pull/155?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical) (https://github.com/canonical/craft-cli/commit/0a3da0f8fac931e36561fc4d972bf385671f8d10) into [main](https://app.codecov.io/gh/canonical/craft-cli/commit/710760ec9ea1634f8e4361d468ec4b2e0dfc20ec?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical) (https://github.com/canonical/craft-cli/commit/710760ec9ea1634f8e4361d468ec4b2e0dfc20ec) will increase coverage by 25.38%.

If only it were always this easy huh

craft_cli/dispatcher.py
```diff
@@ -24,6 +24,9 @@
 from craft_cli.helptexts import HelpBuilder, OutputFormat
 
 
+_SPURIOUS = "DON'T MIND ME I'M JUST A TEST STRING"
```
tigarmo, line 27: reminding myself to remove this

lengau (approved):

PR #158, opened 2023-06-20
docs: document class attributes separately

tigarmo:
Instead of keeping their docs in the main class docstring, use per- attribute docstrings to improve the readibility and "searchability" of the items. Some minor tweaks and adjustments to the text itself, but nothing too major.

Note that we also drop the "sphinx.ext.more_autodoc" extension because it isn't adding anything (that I can tell) other than an annoying double list of init parameters that I couldn't get rid of.

CRAFT-1180

### DIFF
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -29,33 +29,35 @@ class CommandGroup(NamedTuple):
 
     A list of these is what is passed to the ``Dispatcher`` to run commands as part
     of the application.
-
-    :param name: identifier of the command group (to be used in help texts).
-    :param commands: a list of the commands in this group.
     """
 
     name: str
+    """The identifier of the command group (to be used in help texts)."""
+
     commands: Sequence[Type["BaseCommand"]]
+    """A list of the commands belonging in this group."""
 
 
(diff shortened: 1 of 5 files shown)


craft_cli/dispatcher.py
```diff
@@ -93,34 +95,33 @@ class GlobalArgument(NamedTuple):
...
     """Base class to build application commands.
 
-    Subclass this to create a new command; the subclass must define the following attributes:
-
-    - name: the identifier in the command line
-
-    - help_msg: a one line help for user documentation
-
-    - overview: a longer multi-line text with the whole command description
+    Subclass this to create a new command; the subclass must define the ``name``,
```
tigarmo, line 98: Old/New:
image

craft_cli/errors.py
```diff
@@ -24,29 +24,33 @@
 
 
 class CraftError(Exception):
-    """Signal a program error with a lot of information to report.
+    """Signal a program error with a lot of information to report."""
```
tigarmo, line 27: Old/new
image

dboddie (approved):

Issue #159, opened 2023-06-21
Allow to show the current progress in the window titlebar


sergio-costas:
When building a snap with several parts using -v, it's usually not possible to know which part is being built due to the long text output. An idea (with a much-to-be-desired proposal here: https://github.com/snapcore/snapcraft/pull/4217 ) would be to allow to show the progress text in the window titlebar, by using the ANSI codes for that.

A proposal would be to add a new boolean named parameter to the `progress` method, that, when set to True, will not only show the string in the screen, but also set it as the window title.

These are some ideas for that named parameter:

- update_titlebar
- set_titlebar
- set_on_titlebar
- titlebar
- on_titlebar
- copy_on_titlebar
- dup_on_titlebar

sergio-costas:
@tigarmo I made a proposal implementation in https://github.com/canonical/craft-cli/pull/160

PR #160, opened 2023-06-21
messages: allow to set the terminal window title

sergio-costas:
This patch allows to set the text in the title of a terminal window. The idea is to allow snapcraft to set in the title which step is currently doing.

This is a first proposal for https://github.com/canonical/craft-cli/issues/159


craft_cli/messages.py
```diff
@@ -551,19 +551,24 @@ def _get_progress_params(self, permanent: bool):
         return stream, use_timestamp, ephemeral
 
     @_active_guard()
-    def progress(self, text: str, permanent: bool = False) -> None:
+    def progress(self, text: str, permanent: bool = False, update_titlebar: bool = False) -> None:
```
tigarmo, line 554: Thanks for the PR! I wonder if instead of using a new parameter to `Emitter.progress()` we should instead make it a new parameter in `Emitter.init()`, so that every call to `Emitter.progress()` would update the titlebar. Would that be too noisy?

sergio-costas: I think that it isn't a good idea because there are several "progress" messages that don't match for the title. For example:

    emit.progress(
            "Using snapcraft from snap store channel 'latest/stable' in instance "
            "because snap injection is only supported on Linux hosts.",
            permanent=True,
        )

    emit.progress(
                "Refusing to attach specified UA token as system is already attached.",
                permanent=True,
            )

    emit.progress(
                "Unable to determine CPU count; disabling parallel builds",
                permanent=True,
            )

    emit.progress(
                "--with is no longer supported, export the auth to the environment "
                f"variable {store.constants.ENVIRONMENT_STORE_CREDENTIALS!r} instead",
                permanent=True,
            )

and much more.

tigarmo: you are very right!

tests/unit/test_messages_integration.py
```diff
@@ -296,6 +296,32 @@ def test_progress_verbose(capsys, permanent):
...
+@pytest.mark.parametrize("output_is_terminal", [True])
+def test_title_set_in_tty(capsys, monkeypatch):
+    """Show a progress message with update_title flag."""
+    emit = Emitter()
+    emit.init(EmitterMode.BRIEF, "testapp", GREETING)
+    emit.progress("The meaning of life is 42.", update_titlebar=True)
+    emit.ended_ok()
+
+    out, err = capsys.readouterr()
+    assert (out == "\x1b]2;The meaning of life is 42.\x07") is True
```
tigarmo, line 321:
```suggestion
    assert out == "\x1b]2;The meaning of life is 42.\x07"
```

tigarmo (commented):
Looks good! Can you do a companion PR in Snapcraft using this branch so we can manually test out this change a bit?

sergio-costas:
Ok, you can use https://github.com/snapcore/snapcraft/pull/4217 to test this.

sergio-costas:
@tigarmo I forgot to change this from draft to ready. Done.

tigarmo:
Thanks. I'll try this out and give you feedback today

tigarmo:
I did some experimenting and found a couple of problems:

1) The escape code to set the window title doesn't work on default Konsole. It does work on xterm. Which terminal are you using? I can get it to work on Konsole if I change the "Tab title format" option to `%w`:
image
I wonder if there's a "more portable" way to do this.

2) The last set title persists after Snapcraft exits. See the end of this gif:
craft-cli-window-title
we should clear that up, ideally restoring the original title but I don't know how feasible that is.

sergio-costas:
I'm using Terminus, which uses VTE.Terminal class inside. The codes that I use in the patch are the XTerm ones, yes... https://tldp.org/HOWTO/Xterm-Title-3.html

Can you test this?

imagen

it seems that it's the ANSI code for Konsole...

About preserving the title... yes, you are right. I didn't implement that because, in my case, it already does that due to a script that I have for GIT, but it should be preserved and restored. I'll try it.

sergio-costas:
I'll try changing the ANSI command from 2 (set the title) to 0 (set the title and the icon) just in case...

sergio-costas:
Sent a bug report to KDE about the Konsole problem: https://bugs.kde.org/show_bug.cgi?id=471353

sergio-costas:
@tigarmo Mmm... odd... in ubuntu there is this line in ~/.bashrc:

imagen

It forces the window title to the current folder every time the prompt appears... which makes sense, because not all programs restore the window title.

Fedora has a similar line:

imagen

Are you using a ubuntu-derived distro? Which value returns in your system `echo $TERM`? (in my case, it returns `xterm-256color`).

Because all this seems to point to the konsole bug that delays the title change.

sergio-costas:
@tigarmo This should be ready for merging. As commented, the konsole bug seems to be a problem in konsole itseld, that doesn't refresh the UI unless there is a change in the terminal itself. Also, it seems to be not possible to read the current title in the terminal (at least in a standard way), since it is set using ANSI commands.

sergio-costas:
Also, the linting bug seems to be a thing of setuptools_scm.

PR #164, opened 2023-07-10
format: allow use "R|" for raw text

syu-w:
Help text starts with "R|" will not be reformatted.

### DIFF
--- a/craft_cli/helptexts.py
+++ b/craft_cli/helptexts.py
@@ -63,7 +63,11 @@ def _build_item_plain(title: str, text: str, title_space: int) -> List[str]:
     # the first 4 spaces, the two spaces to separate title/text, and the ':'
     not_title_space = 7
     text_space = TERMINAL_WIDTH - title_space - not_title_space
-    wrapped_lines = textwrap.wrap(text, text_space)
+    raw_text = text.startswith("R|")
+    if raw_text:
+        wrapped_lines = text[2:].splitlines()
+    else:
+        wrapped_lines = textwrap.wrap(text, text_space)
 
     # first line goes with the title at column 4
     first = f"    {title:>{title_space}s}:  {wrapped_lines[0]}"
@@ -71,7 +75,10 @@ def _build_item_plain(title: str, text: str, title_space: int) -> List[str]:
 
     # the rest (if any) still aligned but without title
     for line in wrapped_lines[1:]:
-        result.append(" " * (title_space + not_title_space) + line)
+        if raw_text:
+            result.append(line)
+        else:
+            result.append(" " * (title_space + not_title_space) + line)
 
     return result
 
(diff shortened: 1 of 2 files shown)


craft_cli/dispatcher.py
```diff
@@ -228,7 +228,11 @@ def load_command(self, app_config: Any) -> BaseCommand:
         self._loaded_command = self._command_class(app_config)
 
         # load and parse the command specific options/params
-        parser = _CustomArgumentParser(self._help_builder, prog=self._loaded_command.name)
+        parser = _CustomArgumentParser(
+            self._help_builder,
+            prog=self._loaded_command.name,
+            formatter_class=argparse.RawTextHelpFormatter,
```
tigarmo, line 234: what does this do exactly?

syu-w: This disable text wrap
```
parser.add_argument(
            ...
            help="""
                1. XXX
                2. YYY
                3. ZZZ
            """
        )
```
to
`1. XXX 2. YYY 3. ZZZ`

when display.

charmcraft want to display a list of profiles with descriptions and links.

tigarmo: I see. But then it sounds like it might break code that is currently relying on the default behavior of trimming spaces and wrapping words right? Maybe this should be a constructor parameter instead, so that client code can opt-in?
Also we should test this

craft_cli/dispatcher.py
```diff
@@ -311,7 +317,12 @@ def _get_requested_help(
 
         # instantiate the command and fill its arguments
         command = cmd_class(None)
-        parser = _CustomArgumentParser(self._help_builder, prog=command.name, add_help=False)
+        parser = _CustomArgumentParser(
```
lengau, line 320: Can we get a test to ensure that the correct formatter gets used?

tigarmo:
@syu-w can you give an example of a charmcraft command line call that you want to make, and how it looks with and without this change? This might inform what we test

syu-w:
This is the long lines:
https://github.com/canonical/charmcraft/pull/1149/files#diff-d39be6c0a6ba463f975f6d8f301c531831ca748e6dc34071278786a6ba26b150R123-R141

I found even after this change it sill not work as expected.

syu-w:
Since craft-cli use it's own way to format text, I changed the way it handles this by recognize "R|" and prevent formatting the corresponding string

craft_cli/helptexts.py
```diff
@@ -353,7 +360,8 @@ def _build_markdown_command_help(self, usage, overview, options, other_command_n
             "|-|-|",
         ]
         for title, text in options:
-            option_lines.append(f"| `{title}` | {text} |")
+            _text = text.replace("\n", "<br>")
```
tigarmo, line 363: what does this do?

craft_cli/helptexts.py
```diff
@@ -353,7 +360,8 @@ def _build_markdown_command_help(self, usage, overview, options, other_command_n
...
+            option_lines.append(f"| `{title}` | {_text} |")
```
tigarmo, line 364: what happens to the `R|` prefix when showing as markdown?

syu-w: Oh, that should be removed.

craft_cli/helptexts.py
syu-w: Markdown not support new line in table.

craft_cli/helptexts.py
tigarmo: Ok, we should test that the `"R|"` prefix gets removed in the markdown case then (that's probably what the coverage is complaining

PR #185, opened 2023-10-10
feat: add secrets API

### LINKED ISSUE #184: Add a "secrets" API
### What needs to get done

This API will let applications define words that are "secrets". When printing, these secrets are replaced with a standard password marker (e.g. `******`).

### Why it needs to get done

This is to support build-secrets in craft applications.

tigarmo:
The public API consists of a single method, `Emitter.set_secrets()`, that receives a list of strings and causes said strings to be masked by "*****" in every output generated by the emitter's printer. This includes regular emitter messages, messages generated by log calls, spinner, progress bar and the "streaming brief" prefix.

Fixes #184

### DIFF
--- a/craft_cli/printer.py
+++ b/craft_cli/printer.py
@@ -168,6 +168,8 @@ class Printer:
 
         self.terminal_prefix = ""
 
+        self.secrets: list[str] = []
+
         # run the spinner supervisor
         self.spinner = _Spinner(self)
         if not TESTMODE:
@@ -359,13 +361,15 @@ class Printer:
         if self.stopped:
             return
 
+        text = self._apply_secrets(text)
+
         msg = _MessageInfo(
             stream=stream,
             text=text.rstrip(),
             ephemeral=ephemeral,
             use_timestamp=use_timestamp,
             end_line=end_line,
-            terminal_prefix=self.terminal_prefix,
+            terminal_prefix=self._apply_secrets(self.terminal_prefix),
         )
         self._show(msg)
         if not avoid_logging:
@@ -381,6 +385,8 @@ class Printer:
         use_timestamp: bool,
     ) -> None:
         """Show a progress bar to the given stream."""
+        text = self._apply_secrets(text)
+
         msg = _MessageInfo(
             stream=stream,
             text=text.rstrip(),
@@ -416,3 +422,12 @@ class Printer:
 
         self.log.close()
         self.stopped = True
+
+    def set_secrets(self, secrets: list[str]) -> None:
+        """Set the list of strings that should be masked out in all outputs."""
+        self.secrets = secrets[:]
+
+    def _apply_secrets(self, text: str) -> str:
+        for secret in self.secrets:
+            text = text.replace(secret, "*****")
+        return text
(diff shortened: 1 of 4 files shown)


craft_cli/printer.py
```diff
@@ -416,3 +422,12 @@ def stop(self) -> None:
 
         self.log.close()
         self.stopped = True
+
+    def set_secrets(self, secrets: list[str]) -> None:
+        """Set the list of strings that should be masked out in all outputs."""
+        self.secrets = secrets[:]
```
sergiusens, line 428: what's the objective of the `[:]`? do we need a method instead of just setting the attrib?

tests/integration/test_messages_integration.py
```diff
@@ -1437,3 +1437,52 @@ def test_streaming_brief_spinner(capsys, logger, monkeypatch, init_emitter):
         Line("Done stage"),
     ]
     assert_outputs(capsys, emit, expected_err=expected_err, expected_log=expected_log)
+
+
+@pytest.mark.parametrize("output_is_terminal", [True])
+def test_secrets_integrated(capsys, logger, monkeypatch, init_emitter):
```
sergiusens, line 1443: to document the API expectations, it might be worth to also do a "with pause" and print instead of emit.

craft_cli/printer.py
tigarmo: This is to avoid client code modifying the list by accident after the fact, tested [here](https://github.com/canonical/craft-cli/pull/185/files#diff-026c46a62cfc6af61d84f79ea88769a17b59fb54e31346f67fdcc936918afd16R1154)
I suppose `secrets.copy()` would've been clearer, guess old habits die hard

tigarmo: updated

tests/integration/test_messages_integration.py
tigarmo: done; I actually found an issue with the test so thanks for pointing it to me!

sergiusens (approved):

craft_cli/printer.py
sergiusens: I would have though this would have only been a problem if the method had a default set to a list (where the common practice was to set to None to avoid this)

craft_cli/printer.py
```diff
@@ -416,3 +422,13 @@ def stop(self) -> None:
 
         self.log.close()
         self.stopped = True
+
+    def set_secrets(self, secrets: list[str]) -> None:
+        """Set the list of strings that should be masked out in all outputs."""
+        # Keep a copy, to project against clients modifying the list on accident.
```
cmatsuoka, line 428: Is this intended to be "protect"?

tests/unit/test_printer.py
```diff
@@ -1123,3 +1123,100 @@ def test_spinner_silent_on_complete_messages(spinner, monkeypatch):
     time.sleep(0.05)
 
     assert spinner.printer.spinned == []
+
+
+def test_secrets(capsys, log_filepath):
```
cmatsuoka, line 1128: It could be interesting to add tests for secrets not matching word boundaries (e.g. "term" and "range" to get "apple banana o***** wa*****elon"), multiple repetitions of the secret in the same string, or non-ASCII unicode characters.

tigarmo: these are all very good suggestions, thanks! Added some more tests

cmatsuoka (approved):
Thanks!

PR #196, opened 2023-10-24
feat: allow the application to choose the help order of commands

### LINKED ISSUE #195: Allow the application to choose the order of commands within a command group
### What needs to get done

Provide a command group configuration that keeps the order of the command names in the created group rather than alphabetising them.

### Why it needs to get done

For most command groups, alphabetical by command name is completely sensible. However, some command groups, such as lifecycle commands, have a distinct order. In these cases, we should be able to retain the order.

lengau:
fixes #195

### DIFF
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -40,7 +40,7 @@ class CommandGroup(NamedTuple):
     """A list of the commands belonging in this group."""
 
     ordered: bool = False
-    """Whether this command group is already in the correct group (defaults to False)."""
+    """Whether the commands in this group are already in the correct order (defaults to False)."""
 
 
 class GlobalArgument(NamedTuple):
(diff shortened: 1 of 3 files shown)


craft_cli/dispatcher.py
```diff
@@ -39,6 +39,9 @@ class CommandGroup(NamedTuple):
     commands: Sequence[type[BaseCommand]]
     """A list of the commands belonging in this group."""
 
+    ordered: bool = False
+    """Whether this command group is already in the correct group (defaults to False)."""
```
tigarmo, line 43:
```suggestion
    """Whether the commands in this group are already in the correct order (defaults to False)."""
```

lengau: That's what I get for typing a docstring during huddle

tigarmo (approved):
thanks!

PR #203, opened 2023-11-16
fix: clear prefix for message() and error()

### LINKED ISSUE #202: In "streaming brief" mode, `emit.message()` should clean the progress prefix
### Bug Description

If I call `emit.progress("Hi")` and then `emit.message("Done!")`, the text that is actually shown is `Hi :: Done!`.

### To Reproduce

Something like:

```
emit.progress("Hi", permanent=False)
emit.message("Done!")
```

### part yaml

```shell
n/a
```

### Relevant log output

```shell
n/a
```

tigarmo:
The first commit fixes a bug I bumped into where an `emit.progress()` followed by an `emit.message()` would send the carriage return (to clear the ephemeral progress) to the wrong stream. The second commit then does the actual clearing of the prefix.

Fixes #202

### DIFF
--- a/tests/integration/test_messages_integration.py
+++ b/tests/integration/test_messages_integration.py
@@ -1307,6 +1307,34 @@ def test_capture_delays(tmp_path, loops, sleep, max_repetitions):
 
 
 @pytest.mark.parametrize("output_is_terminal", [True])
+def test_progress_and_message(capsys, logger):
+    emit = Emitter()
+    emit.init(EmitterMode.BRIEF, "testapp", GREETING)
+    emit.progress("The meaning of life is 42.")
+    emit.progress("Another message.")
+    emit.message("Finished successfully.")
+    emit.ended_ok()
+
+    expected_term = [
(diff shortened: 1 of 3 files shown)


tests/integration/test_messages_integration.py
```diff
@@ -1306,6 +1306,34 @@ def test_capture_delays(tmp_path, loops, sleep, max_repetitions):
...
+    emit.progress("The meaning of life is 42.")
+    emit.progress("Another message.")
+    emit.message("Finished successfully.")
+    emit.ended_ok()
+
+    expected_term = [
+        Line("The meaning of life is 42.", permanent=False),
+        Line("Another message.", permanent=False),
+    ]
+    expected_out = [Line("Finished successfully.", permanent=True)]
```
tigarmo, line 1322: @sergiusens without the fix, this line would've been `"Another message. :: Finished successfully."`

mr-cal (approved):

PR #220, opened 2024-01-12
feat(dispatcher): constrained global arguments

### LINKED ISSUE #219: Provide limited choices for global arguments
### What needs to get done

Global arguments should have a way to provide a limited set of choices or a type coercion for the application.

Two examples are the built-in `--verbosity`, which has a limited set of choices, and charmcraft's `--project-dir`, which provides a specific type:

```python
verbosity = GlobalArgument(
    "verbosity",
    "option",
    None,
    "--verbosity",
    "Set the verbosity level to 'quiet', 'brief', 'verbose', 'debug' or 'trace'",
    choices=[mode.name.lower() for mode in EmitterMode],
)
project_dir = GlobalArgument(
    "project_dir",
    "option",
    None,
    "--project_dir",
    "Set the base directory for the project.",
    type=pydantic.DirectoryPath,
)
```

These names are chosen to match the `choices` and `type` of an argument in `argparse`

### Why it needs to get done

Right now an app needs to parse and provide error messages for this by itself, but `craft-cli` can often do this itself.

lengau:
Fixes #219

### DIFF
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -18,10 +18,11 @@
 from __future__ import annotations
 
 import argparse
+import dataclasses
 import difflib
-from typing import Any, Literal, NamedTuple, NoReturn, Optional, Sequence
+from typing import Any, Callable, Literal, NamedTuple, NoReturn, Optional, Sequence
 
-from craft_cli import EmitterMode, emit
+from craft_cli import EmitterMode, emit, utils
 from craft_cli.errors import ArgumentParsingError, ProvideHelpException
 from craft_cli.helptexts import HelpBuilder, OutputFormat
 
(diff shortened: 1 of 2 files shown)


craft_cli/dispatcher.py
```diff
@@ -93,6 +116,9 @@ class GlobalArgument(NamedTuple):
         None,
         "--verbosity",
         "Set the verbosity level to 'quiet', 'brief', 'verbose', 'debug' or 'trace'",
+        choices=[mode.name.lower() for mode in EmitterMode],
+        validator=lambda mode: EmitterMode[mode.upper()],
+        case_sensitive=False,
```
cmatsuoka, line 121: `choices` also exists in argparse, will `validator` and `case_sensitive` options increase the gap between global argument handling and command-specific argument handling instead of reducing it? Ideally we should have an uniform way to handle all arguments.

lengau: `validator` is equivalent to argparse's `type` - I can rename if you'd like.

`case_sensitive` is different, but is there for backwards compatibility in the way we handle `--verbosity`, which was previously case-insensitive due to special handling. Looking over it again, I wonder if I could just run the validator before limiting choices and make a special `verbosity_validator` function that returns the correct emitter mode.

PR #240, opened 2024-02-15
fix(dispatcher): explicitly disable exception chaining

lengau:
In these cases, the chaining doesn't provide useful additional data.

Drive-by: updates ruff to 0.2

### DIFF
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -311,7 +311,7 @@ class Dispatcher:
         except KeyError:
             allowed = (repr(of.name) for of in OutputFormat)
             msg = f"Invalid value for --format; allowed are: {', '.join(sorted(allowed))}"
-            raise self._build_usage_exc(msg)  # noqa: TRY200 (Use `raise from`)
+            raise self._build_usage_exc(msg) from None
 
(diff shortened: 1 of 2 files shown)


mr-cal (approved):
LGTM (I didn't test but I am assuming the traceback will look the same)

lengau:
@mr-cal This does change the traceback, since it sets the `__context__` to `None` rather than the previous exception. However, the context exceptions aren't useful to either an end user or a developer using this library since they're implementation details (they're mostly `KeyError` in places where we use dictionaries and in one case a `StopIteration` where we call `next()`). In all cases, these are raising `ArgumentParsingError`, which the application should always catch and display (but should never care about context).

mr-cal (approved):
> @mr-cal This does change the traceback, since it sets the `__context__` to `None` rather than the previous exception. However, the context exceptions aren't useful to either an end user or a developer using this library since they're implementation details (they're mostly `KeyError` in places where we use dictionaries and in one case a `StopIteration` where we call `next()`). In all cases, these are raising `ArgumentParsingError`, which the application should always catch and display (but should never care about context).

That sounds reasonable to me, thanks for the explanation!

tigarmo (approved):

PR #257, opened 2024-06-21
feat: split the docs url into base and slug

tigarmo:
This commit provides a way to separate the base url for the project's docs and the slug for a specific topic. In particular, this is useful when raising a CraftError so that it holds just the slug of the user documentation for that issue, which is then combined by the Emitter into a full url when reporting the error to the user.


sergiusens (approved):
great! I just noticed we are doing multiple calls to printer instead of one, which we probably want to fix, but not today!

mr-cal (approved):

PR #277, opened 2024-09-26
fix: log arguments with debug

mr-cal:

Parsed arguments can be helpful when debugging.

### DIFF
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -245,7 +245,7 @@ class Dispatcher:
         parser = _CustomArgumentParser(self._help_builder, prog=self._loaded_command.name)
         self._loaded_command.fill_parser(parser)
         self._parsed_command_args = parser.parse_args(self._command_args)
-        emit.trace(f"Command parsed sysargs: {self._parsed_command_args}")
+        emit.debug(f"Command parsed sysargs: {self._parsed_command_args}")
         return self._loaded_command
 
     def parsed_args(self) -> argparse.Namespace:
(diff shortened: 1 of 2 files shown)


craft_cli/dispatcher.py
```diff
@@ -245,7 +245,7 @@ def load_command(self, app_config: Any) -> BaseCommand:
         parser = _CustomArgumentParser(self._help_builder, prog=self._loaded_command.name)
         self._loaded_command.fill_parser(parser)
         self._parsed_command_args = parser.parse_args(self._command_args)
-        emit.trace(f"Command parsed sysargs: {self._parsed_command_args}")
+        emit.debug(f"Command parsed sysargs: {self._parsed_command_args}")
```
tigarmo, line 248: I know that I was the one that suggested this but now I'm afraid of leaking stuff that we shouldn't leak from the command line. For example, what if someone calls `snapcraft pack  --ua-token <token>`?

lengau: This is a general problem with CLI stuff. It gets logged in your bash history too.

Secrets like that really should be passed through environment variables.

lengau (approved):
Wouldn't this be increasing the logging rather than reducing? 🤓

mr-cal:
> Wouldn't this be increasing the logging rather than reducing? 🤓

Ha yes.  I was thinking reducing the log level, which does increase the amount of logging.

mr-cal:
As discussed with @tigarmo and @lengau, we need a solution to hide secrets before logging command line arguements.

PR #295, opened 2024-10-21
feat: support app_config in pre_parse_args()

### LINKED ISSUE #292: Pass application context when generating help messages
### What needs to get done

Pass metadata here when generating help messages:

https://github.com/canonical/craft-cli/blob/faa05d0c1fc53fdc627d10dbbd513eb78c7e9738/craft_cli/dispatcher.py#L335

### Why it needs to get done

App metadata, such as the app name, can be accessed when running the command but not when showing a help message for the command.

A very simple reproducer is:
```python
parser.add_argument(
    "--option",
    type=str,
    help=f"Option to use for {self._app.name}",
)

```

tigarmo:
This lets the commands of a given application access the configuration dict during fill_parser() even when building the help message, like in `app cmd -h`. Of course, this can only be used in cases where the configuration dict contains data that the application already has at pre_parse_args()-time.

Fixes #292

### DIFF
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -281,7 +281,7 @@ class Dispatcher:
         return ArgumentParsingError(self._help_builder.get_usage_message(text))
 
     def _get_requested_help(  # noqa: PLR0912 (too many branches)
-        self, parameters: list[str]
+        self, parameters: list[str], app_config: Any
     ) -> str:
         """Produce the requested help depending on the rest of the command line params."""
         if len(parameters) == 0:
@@ -332,7 +332,7 @@ class Dispatcher:
             raise self._build_usage_exc(msg) from None
 
         # instantiate the command and fill its arguments
-        command = cmd_class(None)
+        command = cmd_class(app_config)
         parser = _CustomArgumentParser(self._help_builder, prog=command.name, add_help=False)
         command.fill_parser(parser)
 
@@ -413,7 +413,7 @@ class Dispatcher:
                 filtered_sysargs.append(sysarg)
         return global_args, filtered_sysargs
 
-    def pre_parse_args(self, sysargs: list[str]) -> dict[str, Any]:
+    def pre_parse_args(self, sysargs: list[str], app_config: Any = None) -> dict[str, Any]:
         """Pre-parse sys args.
 
         Several steps:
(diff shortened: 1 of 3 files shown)


tests/unit/test_help.py
```diff
@@ -1458,3 +1461,31 @@ def test_helprequested_command_format_truncated(parameters):
...
+
+        choices = self.config["choices"]
+        parser.add_argument(
+            "number",
+            choices=choices,
+            help=f"The number to use. Possible values are {choices}.",
+        )
+
+
+def test_helprequested_command_app_config():
```
upils, line 1483: suggestion: I understand this is only testing the `if global_args["help"]` branch in `pre_parse_args()`. I suggest adding one more test for the `if global_args["help"]` branch.

 Maybe convert this test to a parametrized one and provide `--help` and `help` as argument?

craft_cli/dispatcher.py
```diff
@@ -281,7 +281,7 @@ def _build_usage_exc(self, text: str) -> ArgumentParsingError:
         return ArgumentParsingError(self._help_builder.get_usage_message(text))
 
     def _get_requested_help(  # noqa: PLR0912 (too many branches)
```
upils, line 283: suggestion: The signature changed and giving a not None value is only indirectly tested (while testing `dispatcher.pre_parse_args()`). 

Maybe add at least one test in `test_help.py` to make sure `_get_requested_help()` is behaving as expected with some values (especially not a dict, that will make the command unhappy)?

upils (commented):
LGTM. I suggested a couple of improvements in tests.

tests/unit/test_help.py
tigarmo: good catch! Updated following you suggestion, thanks

craft_cli/dispatcher.py
tigarmo: I added a new test that explicitly doesn't pass the second parameter to `pre_parse_args()`, thanks for pointing this out.

In my opinion this module already tests too much of the internals. It's a problem in `craft-cli` as a whole so now I'm trying to stick to the public interfaces and the expected, user-visible outcomes.

mr-cal (approved):
Thanks!

upils (approved):
LGTM, thanks!

PR #301, opened 2024-11-08
fix: only greet once when setting the same mode

tigarmo:
With this fix, ``Emitter.set_mode(mode)`` becomes a no-op if the emitter is already at that ``mode``. This scenario can happen, for example, if the Emitter is initialized with a mode and then the Dispatcher sets the same mode during command-line handling.


bepri (approved):

dariuszd21 (approved):
Thank you!

PR #306, opened 2024-12-05
feat: remove code backticks in CLI help messages

### LINKED ISSUE #303: Remove backticks from generated help messages
### What needs to get done

Parse out the extra backticks from the auto-generated help messages.

### Why it needs to get done

Currently, generated help messages preserve the double backticks (``) from reST documentation when printing help messages to the terminal. These look out of place and muddy up the appearance of the help messages.

bepri:

Closes #303.

(CRAFT-3675)

### DIFF
--- a/craft_cli/helptexts.py
+++ b/craft_cli/helptexts.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 
 import argparse
 import enum
+import re
 import textwrap
 from operator import attrgetter
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


craft_cli/helptexts.py
```diff
@@ -325,6 +326,9 @@ def _build_plain_command_help(
         )
 
         overview = textwrap.indent(command.overview, "    ")
+        # Remove reST-style double backticks
+        # Match _only_ double backticks, never triples
+        overview = re.sub(r"(?<!`)``(?!`)", "", overview)
```
lengau, line 331: Normally I would suggest that we `re.compile` this into a global constant, but I think in this case this method should only ever run once per process invocation so it's probably not worth it?

lengau (approved):
Nice, thanks!

craft_cli/helptexts.py
bepri: I think it actually might make a separate call to this function for each field it's validating, so I'll move it to a constant. Good catch!

tigarmo (approved):
thanks!

PR #309, opened 2024-12-12
fix: hide positional arguments with no help

### LINKED ISSUE #308: Craft-cli crashes when rendering positional arguments in help messages
### What needs to get done

Do not crash if a positional argument has no help text [here](https://github.com/canonical/craft-cli/blob/f547a04ff19372c7747a41cc3ded45731d4e5598/craft_cli/helptexts.py#L72).

### Why it needs to get done

Craft-cli 2.8.0 [added support](https://github.com/canonical/craft-cli/pull/271/files#diff-2742d7ab39b1a86f8f7e964667b45b35bcfefba07335139c2bcabdd6e50d45eaR315) for rendering positional arguments in help messages.  If the argument does not have help text, craft-cli crashes.

source: https://github.com/canonical/snapcraft/issues/5169

sergiusens:
Keep the parsing so they show in the command syntax.

Fixes #308


tigarmo (approved):

craft_cli/helptexts.py
```diff
@@ -336,10 +337,12 @@ def _build_plain_command_help(
 
         if parameters:
             # command positional arguments
-            positional_args_lines = ["Positional arguments:"]
+            positional_args_lines: list[str] = []
             for title, text in parameters:
```
dariuszd21, line 341: Shouldn't you check here if `text is not None` as well or `_build_item_plain` already handles that?

dariuszd21 (approved):
Thank you!

craft_cli/helptexts.py
sergiusens: _build_item_plain handles that.

tigarmo (changes requested):
can you add a changelog entry please

tigarmo (approved):
thanksss

Issue #162, opened 2023-06-30
pytest plugin `assert_error`


lengau:
### What needs to get done

Like the various message asserts, the pytest plugin's emitter should be able to assert that the `error` method was called and check the corresponding craft error.

### Why it needs to get done

Right now there's no way to check that the correct error was sent to `emit.error`

lengau:
This is getting reverted in https://github.com/canonical/craft-cli/pull/311.

The feature works, but I didn't catch when creating it that it's a breaking change, so we'll need to assign it to the next major release.

PR #311, opened 2024-12-12
fix(pytest): revert "add an assert_error plugin method"

lengau:
Reverts canonical/craft-cli#218

Breaks #162 

Confirmed this is the cause. Compare:

- With #218: https://github.com/canonical/craft-application/actions/runs/12305359436/job/34344765176?pr=586
- Without: https://github.com/canonical/craft-application/actions/runs/12305776816/job/34346114053?pr=594

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,7 +10,8 @@ included in each version.
 2.12.0 (2024-Dec-XX)
 --------------------
 
-- Remove the ``assert_error`` plugin method.
+- Remove the ``assert_error`` pytest plugin method. For checking errors, we
+  recommend using ``capsys`` instead.
 - Add a ``confirm`` method to the emitter for asking a yes-no question.
 
 2.11.0 (2024-Dec-12)
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,11 @@ Changelog
 See the `Releases page`_ on GitHub for a complete list of commits that are
 included in each version.
 
+2.12.0 (2024-Dec-XX)
+--------------------
+
+- Remove the ``assert_error`` plugin method.
+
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -7,6 +7,11 @@ Changelog
 See the `Releases page`_ on GitHub for a complete list of commits that are
 included in each version.
 
+2.12.0 (2024-Dec-XX)
+--------------------
+
+- Remove the ``assert_error`` plugin method.
```
medubelko, line 13: Doing diligence: is there a replacement or fallback?

lengau: No, this is removing it because it was a breaking change.

lengau: Hmm though maybe:

```suggestion
- Remove the ``assert_error`` pytest plugin method. Use ``capsys`` to check for errors.
```

medubelko: The tense makes the grammar a bit ambiguous.

```suggestion
- Remove the ``assert_error`` pytest method. For checking errors, we recommend
  using ``capsys`` instead.
```

sergiusens (approved):
to me it is a 2.11.1, but I digress

tigarmo:
> to me it is a 2.11.1, but I digress

if we want to get pedantic about it this would be a 3.0 actually

tigarmo (approved):

lengau:
@tigarmo if we want to get really pedantic about it the only bug in 2.11 is that it wasn't called 3.0 ;-)

docs/changelog.rst
lengau: Done!

medubelko (approved):

PR #319, opened 2025-01-13
feat: bash auto-completion generation

### LINKED ISSUE #211: Generate bash completion for the application
### What needs to get done

Make a way to generate a shell completion file (even if this doesn't get included in the final application)		

### Why it needs to get done

Manually writing bash completion is a pain.

bepri:

Closes #211.
CRAFT-3854


bepri:
Linting errors appear unrelated to this PR

tests/integration/test_completion.py
```diff
@@ -0,0 +1,149 @@
...
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+import textwrap
+from pathlib import Path
+
+import craft_cli
+from craft_cli.completion import complete
+
+
+class TestLsCommand(craft_cli.BaseCommand):
```
lengau, line 23:
```suggestion
class FakeLsCommand(craft_cli.BaseCommand):
```

Likewise for the cp command - pytest uses the `Test` prefix for classes to presume that the class contains tests, and then it complains.

lengau (changes requested):
The documentation linting issue is because this PR adds the `utils` module. You just need to add `utils` to the sphinx wordlist.

tests/integration/test_completion.py
```diff
@@ -0,0 +1,149 @@
...
+        return craft_cli.Dispatcher(
+            appname="pybash",
+            commands_groups=[basic_group],
+            extra_global_args=[],
+        )
+
+    actual_output = complete("testcraft", _get_dispatcher)
+
+    expected_output = textwrap.dedent("""\
+        #!/usr/bin/env bash
```
tigarmo, line 64: can you put this string in a separate file? Like we do in a bunch of craft-application tests
It makes it a) easier to read the expected bash script and b) easier to read the test itself, without the huge string in the middle

tox.ini
```diff
@@ -58,15 +58,17 @@ commands =
...
+extras = lint, completion
 env_dir = {work_dir}/linting
 runner = ignore_env_name_mismatch
 allowlist_externals =
     ruff: ruff
     shellcheck: bash, xargs
 
 [shellcheck]
-find = git ls-files
+# jinja2 shell script templates are mistakenly counted as "true" shell scripts and create
```
tigarmo, line 69: 👍🏾

craft_cli/completion/completion.py
```diff
@@ -0,0 +1,269 @@
...
+    def from_global_argument(cls, argument: craft_cli.GlobalArgument) -> Self:
+        """Convert a general GlobalArgument to an OptionArgument for parsing."""
+        if argument.short_option:
+            flags = [argument.long_option, argument.short_option]
+        else:
+            flags = [argument.long_option]
+
+        if argument.choices:
+            completion_command = CompGen(words=argument.choices)
+        elif argument.validator == pydantic.DirectoryPath:
```
tigarmo, line 141: this is a very surprising comparison to me. What does craft-cli's GlobalArgument have to do with pydantic?

craft_cli/completion/completion.py
```diff
@@ -0,0 +1,269 @@
...
+import argparse
+import dataclasses
+import enum
+import importlib
+import shlex
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING, Callable, List, Optional, Union, cast
+
+if TYPE_CHECKING:
```
tigarmo, line 27: this check seems less useful since the code is always importing `typing`

craft_cli/completion/completion.py
```diff
@@ -0,0 +1,269 @@
...
+        "func",
+        type=_validate_dispatch_func,
+        metavar="FUNC_REF",
+        help="A reference to a Python function to be imported. The function should take no arguments and return a craft-cli dispatcher.\n"
+        "Example: some.python.module:get_dispatcher",
+    )
+    args = parser.parse_args(sys.argv[1:])
+
+    # Necessary to avoid errors from running foreign functions that use the craft-cli emitter
+    craft_cli.emit.init(
```
tigarmo, line 265: you should call `emit.ended_ok()` at the end

craft_cli/completion/bash_completion.sh.j2
```diff
@@ -0,0 +1,69 @@
...
+      {% for option in options[0] %}
+        {{ option.flags | join("|") }})
+          COMPREPLY=($({{option.completion_command}} -- "$cur"))
+          return
+          ;;
+      {% endfor %}
+        *)
+          # Not in the middle of a command option, present all options.
+          COMPREPLY=(
+            $(compgen -W "{{options[1] | map(attribute="flags") | map("join", " ") | join(" ")}}" -- "$cur")
```
tigarmo, line 47: I'm not opposed on principle of using Jinja for this template, but I think it might make the code easier to maintain (and test) if have as little as possible code on the templates 
how about this: instead of a tuple, can `options` be a smarter object that has methods/properties that do all of these transformations?

craft_cli/completion/completion.py
```diff
@@ -0,0 +1,269 @@
...
+    user = enum.auto()
+    variable = enum.auto()
+
+
+def _get_set_flags(flags: enum.Flag) -> List[enum.Flag]:
+    return [f for f in flags.__class__ if f & flags == f]
+
+
+@dataclasses.dataclass
+class CompGen:
```
tigarmo, line 86: these helper classes are very good, but we need unit tests for them please

craft_cli/dispatcher.py
```diff
@@ -402,20 +429,33 @@ def _parse_options(  # noqa: PLR0912 (too many branches)
...
                 option, value = sysarg.split("=", 1)
-                arg = arg_per_option[option]
-                if not value:
-                    raise self._build_usage_exc(f"The {arg.name!r} option expects one argument.")
-                global_args[arg.name] = value
             else:
                 filtered_sysargs.append(sysarg)
+                continue
+
+            arg = arg_per_option[option]
```
tigarmo, line 445: these changes to the dispatcher feel a bit out of place. Can you share why they're needed?

craft_cli/dispatcher.py
```diff
@@ -442,15 +482,8 @@ def pre_parse_args(self, sysargs: list[str], app_config: Any = None) -> dict[str
...
-        elif global_args["verbosity"]:
-            try:
-                verbosity_level = EmitterMode[global_args["verbosity"].upper()]
-            except KeyError:
-                raise self._build_usage_exc(
-                    "Bad verbosity level; valid values are "
-                    "'quiet', 'brief', 'verbose', 'debug' and 'trace'."
-                ) from None
-            emit.set_mode(verbosity_level)
+        elif verbosity := global_args["verbosity"]:
```
tigarmo, line 485: here too

tigarmo (changes requested):

craft_cli/completion/completion.py
bepri: This is mostly to handle a generic case. If an argument's validator function happens to be `pydantic.DirectoryPath`, we use the directory flag for `compgen` so that tab completion only shows current directories, like with `cd`. I'm not sure if any craft tools do anything like this now. As an example though, if we ever had a global argument like `--log-output` (somewhat silly example, I realize) that takes a directory and outputs many different logs into it, it would make sense to validate it with `pydantic.DirectoryPath`.

craft_cli/completion/completion.py
bepri: This block was necessary for the [TC003 lint](https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/)

craft_cli/dispatcher.py
bepri: I may have been a bit overzealous :sweat_smile:. When adding my own smaller changes to the dispatcher, I saw a lot of places where the path of execution could be simplified and a bit more easy to follow, which also helped me in debugging this initially.

If you'd prefer, I can revert as much as possible.

tigarmo: yes please. This sort of change has a way of breaking downstream in surprising ways

tests/integration/test_completion.py
bepri: Fixing this has the convenient side-effect of being tested by shellcheck, which I wasn't sure how I wanted to accomplish before. Nice!

bepri:
Snapcraft branch passing using this branch: https://github.com/canonical/snapcraft/pull/5208

craft_cli/completion/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""Bash completion script generation module."""
+
+from .completion import (
+    complete,
+    get_set_flags,
```
tigarmo, line 5: do we want to expose all these names? shouldn't clients just use `complete`?

tests/integration/test_completion.py
```diff
@@ -0,0 +1,65 @@
...
+
+        return craft_cli.Dispatcher(
+            appname="pybash",
+            commands_groups=[basic_group],
+            extra_global_args=[],
+        )
+
+    actual_output = complete("testcraft", _get_dispatcher)
+
+    expected_output = (Path(__file__).parent / "test_completion" / "expected_script.sh").read_text()
```
tigarmo, line 63: much better 👍🏾

tests/unit/test_completion.py
tigarmo, line 1: 👍🏾

tigarmo (approved):
very good!

craft_cli/completion/__init__.py
bepri: Clients only use `complete`, but I had to expose them for testing. Is there another mechanism for doing that that I could use?

tigarmo: I suppose the tests could import directly from `craft_cli.completion.completion`

sergiusens: minimizing exposure seems like a good thing to do

bepri: done, thanks!

lengau (approved):
🚀 🚢

PR #320, opened 2025-01-17
chore: bump minimum python version to 3.10

bepri:

### DIFF
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -28,7 +28,7 @@ from craft_cli.helptexts import HelpBuilder, OutputFormat
 from craft_cli.utils import humanize_list
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Sequence
+    from collections.abc import Callable
 
 
 class CommandGroup(NamedTuple):
@@ -41,7 +41,7 @@ class CommandGroup(NamedTuple):
     name: str
     """The identifier of the command group (to be used in help texts)."""
 
-    commands: Sequence[type[BaseCommand]]
+    commands: list[type[BaseCommand]]
     """A list of the commands belonging in this group."""
 
     ordered: bool = False
--- a/craft_cli/dispatcher.py
+++ b/craft_cli/dispatcher.py
@@ -19,14 +19,11 @@ from __future__ import annotations
 
 import argparse
 import difflib
-from typing import TYPE_CHECKING, Any, Literal, NamedTuple, NoReturn, Optional
+from typing import Any, Literal, NamedTuple, NoReturn, Optional
 
 from craft_cli import EmitterMode, emit
 from craft_cli.errors import ArgumentParsingError, ProvideHelpException
 from craft_cli.helptexts import HelpBuilder, OutputFormat
-
-if TYPE_CHECKING:
-    from collections.abc import Sequence
 
 
 class CommandGroup(NamedTuple):
@@ -39,7 +36,7 @@ class CommandGroup(NamedTuple):
     name: str
     """The identifier of the command group (to be used in help texts)."""
 
-    commands: Sequence[type[BaseCommand]]
+    commands: list[type[BaseCommand]]
     """A list of the commands belonging in this group."""
 
     ordered: bool = False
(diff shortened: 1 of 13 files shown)


lengau:
This probably requires a major release.

craft_cli/dispatcher.py
```diff
@@ -20,13 +20,16 @@
...
-from typing import Any, Callable, Literal, NamedTuple, NoReturn, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Literal, NamedTuple, NoReturn, Optional
 
 from craft_cli import EmitterMode, emit
 from craft_cli.errors import ArgumentParsingError, ProvideHelpException
 from craft_cli.helptexts import HelpBuilder, OutputFormat
 from craft_cli.utils import humanize_list
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
```
tigarmo, line 31: I think we can keep this out of the `TYPE_CHECKING` guard. We're already importing `typing` anyway

craft_cli/dispatcher.py
```diff
@@ -38,7 +41,7 @@ class CommandGroup(NamedTuple):
     name: str
     """The identifier of the command group (to be used in help texts)."""
 
-    commands: Sequence[type[BaseCommand]]
+    commands: list[type[BaseCommand]]
```
tigarmo, line 44: This is a semantic change (Sequences are immutable). Any particular reason to change to `list` here?

craft_cli/printer.py
```diff
@@ -29,10 +29,11 @@
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any, Callable, TextIO
+from typing import TYPE_CHECKING, Any, TextIO
 
 if TYPE_CHECKING:
     import pathlib
+    from collections.abc import Callable
```
tigarmo, line 36: same comment here
in fact I think both imports should be top-level

tigarmo (changes requested):

craft_cli/dispatcher.py
tigarmo: I think adding a noqa is fine, you can add a comment saying that the TYPE_CHECKING guard breaks sphinx doc generation
or just disable the ruff rule for craft-cli (probably better)

craft_cli/dispatcher.py
tigarmo: I think we should probably disable (on pyproject.toml) the ruff rule that's being this picky. Should be ok to do this on craft-cli only, because the reason is sphinx autodoc generation which most other projects don't use
I guess what I really want is for ruff to allow "standard" imports like `collections`

bepri: I agree with that last sentence. I don't really know how to achieve that though, so I'm going with adding [TC003](https://docs.astral.sh/ruff/rules/typing-only-standard-library-import/) to the ignored lints.

upils (changes requested):

craft_cli/utils.py
```diff
@@ -15,7 +15,7 @@
 
 """Utility functions for craft_cli."""
 
-from typing import Sequence
+from collections.abc import Sequence
 
 
 def humanize_list(values: Sequence[str], conjunction: str = "and") -> str:
```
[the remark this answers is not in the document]
bepri: It was not, thanks! I missed this usage of `Sequence` (the change above was auto from ruff)

tests/integration/test_completion.py
```diff
@@ -48,7 +48,7 @@ def fill_parser(self, parser: craft_cli.dispatcher._CustomArgumentParser) -> Non
         parser.add_argument("src", type=Path)
         parser.add_argument("dest", type=Path)
 
-def get_app_info_func(commands: Sequence[Type[craft_cli.BaseCommand]], config: Dict[str, Any] = {}) -> Callable[[], DispatcherAndConfig]:
+def get_app_info_func(commands: list[Type[craft_cli.BaseCommand]], config: Dict[str, Any] = {}) -> Callable[[], DispatcherAndConfig]:
```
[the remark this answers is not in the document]
bepri: Huh, yeah it definitely should. I'm not sure why ruff missed this one at all since `Dict` should be semantically identical to `dict` from what I can tell. Thanks

tigarmo (approved):

PR #323, opened 2025-01-22
fix(completion): use an app_config if present

bepri:

Fixes an issue when using the completion generator with craft tools based on craft-application.

CRAFT-3994

### DIFF
--- a/craft_cli/completion/completion.py
+++ b/craft_cli/completion/completion.py
@@ -23,7 +23,7 @@ import shlex
 import sys
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import TYPE_CHECKING, Callable, List, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 import jinja2
 from overrides import override
@@ -31,7 +31,6 @@ from typing_extensions import Self
 
 if TYPE_CHECKING:
     from collections.abc import MutableSequence
-    from typing import Dict
 
 import craft_cli
 
@@ -193,14 +192,16 @@ class CommandMapping:
         return " ".join([a.flag_list for a in self.args])
 
 
-def complete(shell_cmd: str, get_dispatcher: Callable[[], craft_cli.Dispatcher]) -> str:
+def complete(
+    shell_cmd: str, get_dispatcher: Callable[[], Tuple[craft_cli.Dispatcher, Dict[str, Any]]]
+) -> str:
     """Generate a bash completion script based on a craft-cli dispatcher.
 
     :param shell_cmd: The name of the command being completed for
     :param get_dispatcher: A function that returns a populated craft-cli dispatcher
     :return: A bash completion script for ``shell_cmd``
     """
-    dispatcher = get_dispatcher()
+    dispatcher, app_config = get_dispatcher()
     env = jinja2.Environment(
         trim_blocks=True,
         lstrip_blocks=True,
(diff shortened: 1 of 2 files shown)


bepri:
With this branch I'm able to successfully use the newer craft-application based Snapcraft dispatcher: https://github.com/canonical/snapcraft/pull/5208

craft_cli/completion/completion.py
```diff
@@ -215,7 +216,7 @@ def complete(shell_cmd: str, get_dispatcher: Callable[[], craft_cli.Dispatcher])
     command_map: Dict[str, CommandMapping] = {}
     for name, cmd_cls in dispatcher.commands.items():
         parser = argparse.ArgumentParser()
-        cmd = cmd_cls(None)
+        cmd = cmd_cls(app_config)
```
lengau, line 219: Thanks for including me here - this was the piece where I wasn't understanding the need, as I thought the dispatcher had the app config. But refreshing based on the [craft-application implementation](https://github.com/canonical/craft-application/blob/657d1fe9497c0e9c5c2a4d237b25b9b666080eeb/craft_application/application.py#L623) it makes more sense.

lengau (approved):

craft_cli/completion/completion.py
```diff
@@ -193,14 +192,16 @@ def all_args(self) -> str:
         return " ".join([a.flag_list for a in self.args])
 
 
-def complete(shell_cmd: str, get_dispatcher: Callable[[], craft_cli.Dispatcher]) -> str:
+def complete(
+    shell_cmd: str, get_app_info: Callable[[], Tuple[craft_cli.Dispatcher, Dict[str, Any]]]
```
tigarmo, line 196: I think the second element of the tuple should be `Dict[str, Any] | None`, following what BaseCommand [accepts](https://github.com/canonical/craft-cli/blob/main/craft_cli/dispatcher.py#L158).

tigarmo: since you use this Tuple type in a lot of places it might make sense to create something like a `DispatcherAndConfig` type

craft_cli/completion/completion.py
```diff
@@ -193,14 +192,16 @@ def all_args(self) -> str:
...
+) -> str:
     """Generate a bash completion script based on a craft-cli dispatcher.
 
     :param shell_cmd: The name of the command being completed for
-    :param get_dispatcher: A function that returns a populated craft-cli dispatcher
+    :param get_app_info: A function that returns a populated craft-cli dispatcher and its app config
```
tigarmo, line 201:
```suggestion
    :param get_app_info: A function that returns a populated craft-cli dispatcher and the config needed to create its commands
```

tests/integration/test_completion.py
```diff
@@ -48,18 +49,30 @@ def fill_parser(self, parser: craft_cli.dispatcher._CustomArgumentParser) -> Non
...
+    actual_output = complete("testcraft", app_info_func)
 
     expected_output = (Path(__file__).parent / "test_completion" / "expected_script.sh").read_text()
 
     assert actual_output == expected_output
+
+def test_app_config_used() -> None:
+    app_info_func = get_app_info_func([FakeCpCommand], config={"hello": "world"})
+
+    with patch(__name__ + ".FakeCpCommand.__init__", return_value=None) as complete_mock:
```
tigarmo, line 75: I'm not a big fan of mocking inits if we don't have to, it makes the tests harder to reason about.

How about this instead: Create a separate command that, in its init, adds a key to that `config` dict. Then you can check that the dict has the key after calling `complete`.

tigarmo (changes requested):
thanks for the quick turnaround on this

tigarmo (approved):
thanks!

PR #331, opened 2025-03-19
feat(emitter): add a log dumping function

bepri:

Adds a new function to dump external logs directly into the emitter's log file, bypassing printing to the terminal and giving better customization for log behavior in the future.

It additionally adds a public `log_filepath` read-only property so consumers of the emitter can more easily access and test the contents of the log.

Required by https://github.com/canonical/craft-application/pull/691.

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -781,17 +781,16 @@ class Emitter:
         self._stop()
 
     @_active_guard()
-    def dump_log_contents(self, log_path: pathlib.Path | str, prefix: str = "::") -> None:
-        """Dump the contents of an external log file into the emitter log."""
-        # Open a path object in case we received a string
-        if isinstance(log_path, str):
-            log_path = pathlib.Path(log_path)
+    def append_to_log(self, file: TextIO, prefix: str = ":: ") -> None:
+        """Dump the contents of an external log file into the emitter log.
 
-        with log_path.open() as log_file:
-            for line in log_file:
-                text = f"{prefix} {line}"
-                # Don't set the stream so it only goes to the log file
-                self._printer.show(None, text, use_timestamp=False)
+        :param file: A file I/O object to read from
+        :param prefix: A prefix for every line printed. Defaults to ":: ".
+        """
+        for line in file.readlines():
+            text = f"{prefix}{line}"
+            # Don't set the stream so it only goes to the log file
+            self._printer.show(None, text, use_timestamp=False)
 
     @_active_guard()
     def set_secrets(self, secrets: list[str]) -> None:
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -782,7 +782,7 @@ class Emitter:
 
     @_active_guard()
     def dump_log_contents(self, log_path: pathlib.Path | str, prefix: str = "::") -> None:
-        """Dump the contents of a log file."""
+        """Dump the contents of an external log file into the emitter log."""
         # Open a path object in case we received a string
         if isinstance(log_path, str):
             log_path = pathlib.Path(log_path)
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -780,7 +780,7 @@ class Emitter:
         self._report_error(error)
         self._stop()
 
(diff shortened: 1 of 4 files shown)


tests/unit/test_messages_emitter.py
```diff
@@ -88,6 +87,28 @@ def func(mode, *, greeting="default greeting", **kwargs):
 
         yield func
 
+@pytest.fixture
+def emitter_methods() -> Callable[[Emitter | RecordingEmitter, list[str]], list[Callable[..., Any]]]:
```
bepri, line 91: This fixture was added to more generically generate a list of all the methods on an emitter, as a few tests were doing it manually. It was necessary as the tests assumed all fields of the emitter were callables, but properties cannot be called.

craft_cli/messages.py
```diff
@@ -780,6 +780,19 @@ def error(self, error: errors.CraftError) -> None:
         self._report_error(error)
         self._stop()
 
+    @_active_guard()
+    def dump_log_contents(self, log_path: pathlib.Path | str, prefix: str = "::") -> None:
```
lengau, line 784: I'm going to defer to whatever @tigarmo thinks, but what about a method like:

```python
def append_to_log(self, file: TextIO, prefix: str = ":: ") -> None:
```

That way we can add any file-like object to the log. This way in craft-application we don't need to copy the log file to the outer instance and then read that - instead we can read the file directly from the instance and write it directly to the log.

lengau: Other examples of things we could do include writing directly from network streams to the log (e.g. Launchpad build logs) or directly logging command output to the log for commands that are purely internal.

bepri: I like this as well! If Tiago is game then we can do it.

tigarmo: sure that sounds fine

craft_cli/messages.py
```diff
@@ -780,6 +780,19 @@ def error(self, error: errors.CraftError) -> None:
...
+        """Dump the contents of an external log file into the emitter log."""
+        # Open a path object in case we received a string
+        if isinstance(log_path, str):
+            log_path = pathlib.Path(log_path)
+
+        with log_path.open() as log_file:
+            for line in log_file:
+                text = f"{prefix} {line}"
```
tigarmo, line 792: Alex had another suggestion implicitly (unconsciouly?) on his comment: I don't think we should hardcode a space here between `prefix` and `line`, and the method's default should be `":: "` (with the space). That way consumers can use whatever prefix they want, including no prefix at all with `""`

tests/unit/test_messages_emitter.py
```diff
@@ -181,13 +202,12 @@ def test_init_developer_modes(mode, tmp_path, monkeypatch):
     assert handler.mode == mode
 
 
-@pytest.mark.parametrize("method_name", [x for x in dir(Emitter) if x[0] != "_" and x != "init"])
-def test_needs_init(method_name):
+def test_needs_init(emitter_methods):
```
tigarmo, line 205: A problem with this approach is that now the test is no longer parametrized; so instead of having N tests we have a single test that looks but will stop as soon as an assertion fails.

How about, instead of having a fixture, just a regular function that's called during `parametrize`? Like:
```python
@pytest.mark.parametrize("method_name", emitter_methods(exclude=["init"]))
```

tigarmo (changes requested):

craft_cli/messages.py
```diff
@@ -780,6 +780,18 @@ def error(self, error: errors.CraftError) -> None:
...
         self._stop()
 
+    @_active_guard()
+    def append_to_log(self, file: TextIO, prefix: str = ":: ") -> None:
+        """Dump the contents of an external log file into the emitter log.
+
+        :param file: A file I/O object to read from
+        :param prefix: A prefix for every line printed. Defaults to ":: ".
+        """
+        for line in file.readlines():
```
lengau, line 790:
```suggestion
        for line in file:
```

lengau (approved):
🚀

PR #340, opened 2025-05-02
chore: prepare for starbase merge

bepri:

Lots of papercut fixes separated from the starbase merge to keep the commit history in the PR calm. The merge commit will happen in a separate PR after this gets merged.


craft_cli/__init__.py
```diff
@@ -31,8 +31,12 @@
 # is to break cyclic dependencies
 from .messages import EmitterMode, emit  # isort:skip
 from .dispatcher import BaseCommand, CommandGroup, Dispatcher, GlobalArgument
-from .errors import ArgumentParsingError, CraftError, CraftCommandError, ProvideHelpException
-from .helptexts import HIDDEN  # noqa: F401
```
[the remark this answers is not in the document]
bepri: It seems to have been [removed by `ruff check --fix`](https://github.com/canonical/craft-cli/commit/6477314da141e8f00cb85aa022a0557719c5a67e#diff-acf1c713eccc9197d991b5fe4bfc0d857960897bda68ff296479915116d4ff7dL35). The original line's `noqa` was for [unused imports](https://docs.astral.sh/ruff/rules/unused-import/).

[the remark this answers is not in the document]
bepri: Thanks - I didn't realize this was needed externally, but reading the comments around it I definitely see why now. I put it back

.github/workflows/release-publish.yaml
```diff
@@ -3,16 +3,18 @@ on:
   push:
     tags:
       # These tags should be protected, remember to enable the rule:
-      # https://github.com/canonical/craft-cli/settings/tag_protection
+      # https://github.com/canonical/starbase/settings/tag_protection
```
tigarmo, line 6:
```suggestion
      # https://github.com/canonical/craft-cli/settings/tag_protection
```

This should point to the repo's settings

HOWTO_RELEASE.md
tigarmo, line 1: Maybe we should just drop this whole file? We don't publish from local machines anymore

craft_cli/messages.py
```diff
@@ -201,13 +200,17 @@ def __init__(
...
+
             # parameters: default security, default buffer size, binary mode
             binary_mode = os.O_BINARY
             # ignoring the type of the first parameter below, as documentation allows to use None
             # to make it use a NULL security descriptor:
             #     https://www.markjour.com/docs/pywin32-docs/PySECURITY_ATTRIBUTES.html
-            self.read_pipe, self.write_pipe = win32pipe.FdCreatePipe(None, 0, binary_mode)  # type: ignore[reportGeneralTypeIssues]
+            self.read_pipe, self.write_pipe = win32pipe.FdCreatePipe(
+                None, 0, binary_mode
+            )  # type: ignore[reportGeneralTypeIssues]
```
tigarmo, line 213: possibly this type ignore can be dropped now?

tigarmo (approved):
Thanks a lot! So is the plan to a) rebase-merge this PR and then b) do another PR with a no-op regular merge from starbase?

.readthedocs.yaml
```diff
@@ -5,22 +5,28 @@
...
-build:
-  os: ubuntu-22.04
-  tools:
-    python: "3.12"
-
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   builder: dirhtml
   configuration: docs/conf.py
-  fail_on_warning: false
```
mr-cal, line 18: This was just added to starbase yesterday.

HOWTO_RELEASE.md
mr-cal: heh, this workflow uses `fades`

craft_cli/completion/completion.py
```diff
@@ -248,6 +252,8 @@ def complete(shell_cmd: str, get_app_info: Callable[[], DispatcherAndConfig]) ->
         commands=command_map,
         global_args=dispatcher.global_arguments,
         global_opts=global_opts,
+        # The shebang is initially omitted to avoid it being linted by shellcheck
```
mr-cal, line 255: Was shellcheck doing a proper check of the autocompleter script?

I mean, in the future, should we move this back so the script gets linted by shellcheck?

mr-cal (approved):

HOWTO_RELEASE.md
bepri: I agree. I didn't want to be over-eager since it at least talked about tagging, which was relevant, but I'm always happy to delete a top-level file

craft_cli/completion/completion.py
bepri: Sorta. Shellcheck was configured in this repository to explicitly ignore jinja templates: https://github.com/canonical/craft-cli/blob/e3aa3e5b3905d1e301661d99662d30c0cc9dfdc0/tox.ini#L68

An example of a post-jinja script exists in the integration tests, which is how this is being linted normally. Omitting the shebang just makes `file --mime-type` report it as plain text instead of a shellscript, which avoids it getting passed to shellcheck in the first place.

Ideally this is replaced by `common.mk` upstream being modified to ignore jinja templates. I should've just put it there in the first place though instead of making this hacky fix.

PR #358, opened 2025-07-10
feat(emitter): add warning method

### LINKED ISSUE #256: Add `emit.warning()`
### What needs to get done

Add a new emitter command, `emit.warning()`

### Why it needs to get done

`emit.progress(permanent=True)` is overloaded.  It is the only way to log warnings like deprecation notices.

#### Acceptance criteria

* Warning method on the emitter that outputs (permanently) to stderr, includes the warning in the log, and provides a customisable prefix (Default: “WARNING: ”)  
* Tests for this method  
* At least one example in [examples.py](http://examples.py) that uses progress with permanent=False, progress with permanent=True, and warning.

nryanl:

---
Adds a warning method to the Emitter class.
Resolves #256.

### DIFF
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -572,6 +572,22 @@ class Emitter:
         self._printer.show(stream, text, use_timestamp=use_timestamp)
 
     @_active_guard()
+    def warning(self, text: str, *, prefix: str = "WARNING: ") -> None:
+        """Warning information.
+
+        To show warnings to the user, which are not errors but may
+        indicate that something is not right and the user should pay attention to it.
+        """
+        text = f"{prefix}{text}"
+        if self._mode in (EmitterMode.DEBUG, EmitterMode.TRACE):
+            stream = sys.stderr
+            use_timestamp = True
+        else:
+            stream = sys.stderr
+            use_timestamp = False
+        self._printer.show(stream, text, use_timestamp=use_timestamp)
+
+    @_active_guard()
     def debug(self, text: str) -> None:
         """Debug information.
 
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -573,7 +573,7 @@ class Emitter:
 
     @_active_guard()
     def warning(self, text: str, *, prefix: str = "WARNING: ") -> None:
-        """Warning information.
+        """Show an important warning to the user.
 
         To show warnings to the user, which are not errors but may
         indicate that something is not right and the user should pay attention to it.
--- a/craft_cli/messages.py
+++ b/craft_cli/messages.py
@@ -572,14 +572,19 @@ class Emitter:
         self._printer.show(stream, text, use_timestamp=use_timestamp)
 
     @_active_guard()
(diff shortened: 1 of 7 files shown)


craft_cli/pytest_plugin.py
```diff
@@ -158,6 +158,19 @@ def assert_verbose(
         """
         return self._check(expected_text, "verbose", regex)
 
+    def assert_warning(
```
lengau, line 161: Please add a test for this in [test_pytest_plugin.py](https://github.com/canonical/craft-cli/blob/main/tests/test_pytest_plugin.py)

craft_cli/messages.py
```diff
@@ -571,6 +571,22 @@ def verbose(self, text: str) -> None:
             use_timestamp = True
         self._printer.show(stream, text, use_timestamp=use_timestamp)
 
+    @_active_guard()
+    def warning(self, text: str, *, prefix: str = "WARNING: ") -> None:
```
tigarmo, line 575: I think `WARNING` might be too harsh. How about `Warning`?

Also this parameter needs to be documented in the docstring

craft_cli/pytest_plugin.py
```diff
@@ -158,6 +158,19 @@ def assert_verbose(
...
+        self,
+        expected_text: str,
+        regex: bool = False,  # noqa: FBT001
```
tigarmo, line 164: Since this is a brand new method we can (and imo should) make `regex` a keyword-only argument

tigarmo (changes requested):
Thanks for the contribution!

nryanl:
I've incorporated your feedback. Let me know if you have any more suggestions.

craft_cli/messages.py
lengau: (I'm happy with this change, but also pointing out that the text was described in the bounty :-) )

craft_cli/pytest_plugin.py
lengau: Agreed. We can make the change for the rest of the existing methods in the next major release.

craft_cli/messages.py
```diff
@@ -571,6 +571,27 @@ def verbose(self, text: str) -> None:
...
+    def warning(self, text: str, *, prefix: str = "Warning: ") -> None:
+        """Show an important warning to the user.
+
+        To show warnings to the user, which are not errors but may
+        indicate that something is not right and the user should pay attention to it.
+
+        :param prefix: Display text prepended to warnings, defaults to "Warning: "
+        """
+        text = f"{prefix}{text}"
+        if self._mode == EmitterMode.QUIET:
```
lengau, line 584: I think this would look nicer with a [match statement](https://docs.python.org/3/tutorial/controlflow.html#match-statements), but that's not a requirement at all.

lengau (approved):
Thanks @nryanl!

I'm happy with this with the exception of the discussion about whether to emit the warning in quiet mode. However, that discussion is out of scope for this PR and your change matching @tigarmo's expectation is reasonable.

tigarmo (approved):
looks good! Thanks for your contribution

PR #369, opened 2025-07-31
fix: filter other commands by name instead of type

jonathan-conder:
Previously, a command like craft_appplication.command.TestCommand, which inherits from PackCommand, would omit the parent command in the "see also" section.


jonathan-conder:
I modified an existing test to catch the issue. But I'm not sure it really makes sense to test for this particular issue now that it's fixed. At least I'd be surprised if there's a regression.

PR #411, opened 2026-01-22
feat: oxidize the emitter

bepri:
- [ ] Have you successfully run `make lint && make test`?

---

Oxidizes the emitter class. The [PyO3 docs](https://pyo3.rs/v0.27.2/index.html) and [Maturin docs](https://www.maturin.rs/) may help with reviewing. So far, the only intended breaking change is the removal of the global `emit`. Rust, reasonably so, does not play nicely with globally mutable state or partially initialized variables. If anything else is breaking, it's either because it's not yet implemented or I didn't mean to break it, so definitely ask if you spot anything!

Remaining work:
- Streaming brief
- `Emitter.open_stream`
- External log handling (from `logging`)
- Devise testing strategy
- Incremental progress bars
- Fix project versioning (see below)

Overall structure
---

In order to build with Maturin, the build backend just needed to be switched. This was mostly simple, but Maturin has an unfortunate limitation in that it completely lacks anything like `tool.setuptools_scm` for configuring tag-based dynamic versioning. This can be _mostly_ achieved via environment variables, but for local development builds with uv (`maturin develop --uv`), a version _must_ be present in `pyproject.toml`. For now, the version is hard-coded to `0.0.0+dev`.

The field in `pyproject` setting the module name just makes the compiled `.so` appear with that import path. Otherwise, it would overwrite the Python portion of Craft CLI. This new submodule was named `_rs`.

To test what has been done so far, clone this, update your venv, and then run `uv run maturin develop --uv`. This builds to a `.so`, then installs it as an editable wheel.

Comments will be made on code in this PR for any other higher level design choices.

CRAFT-4970

### DIFF
--- a/src/emitter.rs
+++ b/src/emitter.rs
@@ -352,12 +352,6 @@ impl Emitter {
 
     /// Stop the printing infrastructure and print a final message to see the logs.
     fn finish(&mut self) -> PyResult<()> {
-        let message = Message {
-            text: format!("Full execution log at '{}'", self.log_filepath),
-            model: MessageType::Info,
-            target: Some(Target::Stderr),
-        };
-        self.printer.send(message);
         self.printer.stop()?;
         Ok(())
     }


rust/emitter.rs
```diff
@@ -0,0 +1,367 @@
...
+        unimplemented!()
+    }
+
+    /// Stop gracefully.
+    fn ended_ok(&mut self) -> PyResult<()> {
+        self.finish()
+    }
+}
+
+impl Emitter {
```
bepri: Since this lacks `#[pymethods]`, these are effectively private methods.

rust/emitter.rs
```diff
@@ -0,0 +1,367 @@
...
+            model: MessageType::Info(),
+            target: Target::Stderr,
+        };
+        self.printer.send(message);
+        self.printer.stop()?;
+        Ok(())
+    }
+}
+
+impl Drop for Emitter {
```
bepri: This is how destructors are declared.

rust/craft_cli_utils.rs
lengau: Could we rename this directory `src` to follow the standard rust convention?

Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+members = ["."]
+
+[workspace.dependencies]
+# pyo3 is always needed, but tests additionally need the "auto-initialize" feature. To avoid
+# having two disparate version specifications, it's instead specified here exactly once and then
+# retrieved via "workspace = true".
+pyo3 = "0.27.2"
+
+[dependencies]
+console = "0.16.2"
```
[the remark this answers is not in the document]
lengau: https://github.com/console-rs/console

It (although older versions) is also in Ubuntu going all the way back to Jammy. Realistically we don't need to care about it that far back, but it would be interesting to know id 0.15.7 (in noble) works for us.

https://packages.ubuntu.com/search?keywords=librust-console-dev&searchon=names&suite=all&section=all

Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+indicatif = { version = "0.18.0", features = ["improved_unicode"] }
```
lengau: Same question here — does 0.17.7 work for our use case for now? https://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=librust-indicatif-dev&searchon=names

rust/emitter.rs
```diff
@@ -0,0 +1,367 @@
...
+            greeting,
+        })
+    }
+
+    /// Create a log filepath from the app name as an easy default.
+    #[classmethod]
+    fn log_filepath_from_name(_cls: &Bound<'_, PyType>, app_name: String) -> String {
+        let dirs = xdg::BaseDirectories::with_prefix(app_name);
+        let mut p = dirs
+            .get_data_home()
```
lengau: I believe we actually need [`get_state_home`](https://whitequark.github.io/rust-xdg/xdg/struct.BaseDirectories.html#method.get_state_home) here.

.gitignore
```diff
@@ -149,3 +149,8 @@ dmypy.json
 
 # JetBrains IDEs
 /.idea/
+
+
+# Added by cargo
+
+/target
```
lengau: Should we bring in anything else from the [GitHub rust gitignore template](https://github.com/github/gitignore/blob/main/Rust.gitignore)?

rust/craft_cli_utils.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Convert a collection of values into a string that lists the values.
+    #[pyfunction]
+    #[pyo3(signature = (values, conjunction = "and"))]
+    fn humanize_list(mut values: Vec<String>, conjunction: Option<&str>) -> String {
+        let start = values
+            .drain(..values.len() - 1)
+            .collect::<Vec<String>>()
+            .join(", ");
+
+        let conjunction = conjunction.unwrap_or("and");
```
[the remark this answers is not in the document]
lengau: That comes from the signature in the pyo3 macro on line 14, right? Does that mean pyo3 doesn't automatically add "and" as the value if you call it from Python?

rust/craft_cli_utils.rs
```diff
@@ -0,0 +1,36 @@
...
+#[pymodule(submodule)]
+pub mod utils {
+    use pyo3::{Bound, PyResult, pyfunction, types::PyModule};
+
+    use crate::utils::fix_imports;
+
+    /// Convert a collection of values into a string that lists the values.
+    #[pyfunction]
+    #[pyo3(signature = (values, conjunction = "and"))]
+    fn humanize_list(mut values: Vec<String>, conjunction: Option<&str>) -> String {
```
lengau: Although it was annotated with `list`, it actually acecpted any [Sequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence) of strings ([bugs notwithstanding](https://github.com/canonical/craft-cli/issues/413)).

I don't see a Rust equivalent to this, but massive bonus if you can make it accept any Iterable instead.

A naïve Python implementation would be something like:

```python
result = ""
values_iter = iter(values)
item = next(values_iter)
while True:
	try:
    	next_item = next(values_iter)
	except StopIteration:
		break
	result += f", {item}"
	item, next_item = next_item, next(values_iter)
return result + f", {conjunction} {item}"
```

(This still contains several of the same bugs I filed today)

src/craft_cli_utils.rs
```diff
@@ -0,0 +1,53 @@
...
+    #[pyo3(signature = (values, *, conjunction = "and"))]
+    fn humanize_list(values: Bound<'_, PyAny>, conjunction: &str) -> PyResult<String> {
+        // Check if it's actually iterable at runtime and collect values
+        // This could be dramatically simplified if [PyO3#5757](https://github.com/PyO3/pyo3/issues/5757)
+        // is resolved.
+        let items: Vec<_> = match values.try_iter() {
+            Ok(py_iter) => py_iter
+                .into_iter()
+                .map(|maybe_item| maybe_item.map(|item| item.to_string()))
+                .collect::<Result<_, _>>()?,
```
TheSignPainter98: Just to make sure error types are as consistent as possible, it's a good practice to put a declaration like this in the root of your crate, then to make sure to always import `crate::Result`:
```rust
pub type Result<T, E = Error> = std::result::Result<T, E>;
pub type Error = PyErr; // Tying us to pyo3 doesn't seem too bad, in general though, `anyhow::Error` works well here
```

bepri: Ooooh, it looks like PyO3 has an [`anyhow` feature](https://docs.rs/pyo3/latest/pyo3/anyhow/index.html). I think for now I'll just change this to be a `PyResult` and have a follow-up task for switching to `anyhow`. Thanks.

src/printer.rs
```diff
@@ -0,0 +1,363 @@
...
+    fn progress_bar(&mut self, message: &Message) -> PyResult<()> {
+        unimplemented!()
+    }
+}
+
+impl Drop for InnerPrinter {
+    /// Restore the cursor when releasing control of the terminal.
+    fn drop(&mut self) {
+        self.handle_overwrite().unwrap();
+        self.stdout.show_cursor().unwrap();
```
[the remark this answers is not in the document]
TheSignPainter98: That's quite little surprising, are you sure? I haven't seen this pattern with `.map_or_else(|_| ... L, Ok)` before and it looks very similar to `.and_then`, perhaps it was an issue with the error types?

Just a general note, `.and_then` is usually best-avoided in favour of the `?` operator, this is one of the few situations where it may be pleasant

lengau (approved):

PR #416, opened 2026-02-03
feat: oxidize open_stream

bepri:

---

This is based on (and thus requires) #411. The diff will be pretty noisy until it is merged.

I recommend not reviewing the code directly until I've had an opportunity to clean up the commit history once #411 lands, but this can be tested using the scripts below in the meantime.

This PR adds the `open_stream` method back to the emitter. This required refactoring the `Printer` class to not solely exist as a field on an `Emitter` instance, as the stream handler object needed access too. Now, the `Printer` is a global, static variable that gets initialized upon the first print. For similar reasons, the logging mechanism has been moved into the printer and must be explicitly initialized via the new `Emitter.init_logger()`.

Here is some example code to test the changes:

```py
# testing.py
import subprocess

from craft_cli import EmitterMode
from craft_cli._rs.emitter import Emitter

e = Emitter("blah.log", EmitterMode.TRACE, "example.com", "hello from rust!")
e.init_logger()

with e.open_stream() as stream:
    subprocess.run(["bash", "testing.sh"], stdout=stream, stderr=stream)
```

```bash
# testing.sh
for i in {1..6}; do
  echo $i
  sleep 1
done
```

Known bugs: It seems that trace-level events (and perhaps others?) duplicate the timestamp in logs (tracked in CRAFT 4999). Also, the `text` parameter for `open_stream` is missing until CRAFT 4968 is complete, as otherwise the two tasks would duplicate work to add prefix behaviors.

CRAFT-4964.

### DIFF
--- a/src/emitter.rs
+++ b/src/emitter.rs
@@ -1,6 +1,6 @@
 //! The Emitter class and its associated helpers.
 
-use std::{borrow::Cow, path::PathBuf, thread};
+use std::{path::PathBuf, thread};
 
 use pyo3::{Bound, PyResult, pyclass, pymethods, pymodule, types::PyType};
 
@@ -212,7 +212,7 @@ impl Emitter {
     /// by the next line (unless in verbose or trace mode, or set to permanent).
     #[pyo3(signature = (text, *, permanent = false))]
     fn progress(&mut self, text: &str, mut permanent: bool) -> PyResult<()> {
-        let timestamped = Self::apply_timestamp(text);
+        let timestamped = utils::apply_timestamp(text);
 
         let (maybe_timestamped, target) = match self.verbosity {
             Verbosity::Quiet => {
@@ -270,7 +270,7 @@ impl Emitter {
     #[pyo3(signature = (text, prefix = "Warning: "))]
     fn warning(&mut self, text: &str, prefix: &str) -> PyResult<()> {
         let prefixed = format!("{}{}", prefix, text);
-        let timestamped = Self::apply_timestamp(&prefixed);
+        let timestamped = utils::apply_timestamp(&prefixed);
 
         let (maybe_timestamped, target) = match self.verbosity {
             Verbosity::Quiet => (prefixed.as_str(), None),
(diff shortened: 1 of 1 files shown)


src/streams.rs
```diff
@@ -0,0 +1,257 @@
...
+            )));
+        }
+
+        _ = self.write.take();
+
+        Ok(())
+    }
+
+    /// Get a writable file descriptor for this object.
+    fn fileno(&self) -> PyResult<RawFd> {
```
bepri: This is a "magic method" that `os.read()` calls on the file object you hand it to actually write. In a sense, this makes it "fulfill the `IO[str]` protocol" in Python.

src/streams.rs
```diff
@@ -0,0 +1,257 @@
...
+        events: &mut mio::Events,
+        pipe: &mut io::PipeReader,
+    ) -> PyResult<()> {
+        poll.poll(events, Some(Duration::from_millis(100)))?;
+
+        if events.is_empty() {
+            return Ok(());
+        }
+
+        let num_read = match pipe.read(buf) {
```
bepri: `Read` will only fill up to `buf.len()`, so this is equivalent to something like `read(buf, PIPE_READER_CHUNK_SIZE)`. The only error cases I have to handle are an I/O error, or 0 bytes being read (EOF).

I don't directly close the pipe in the event of EOF because that's instead handled by Python calling `__exit__`, which is cleaner.

src/emitter.rs
```diff
@@ -331,6 +296,34 @@ impl Emitter {
...
+    ///
+    /// All messages sent by the emitter will also be sent to this log file moving forward.
+    fn init_logger(&self) -> PyResult<()> {
+        crate::printer::printer().init_logger(&self.log_filepath, &self.greeting)
+    }
+
+    /// Open a stream context manager to redirect output to a different stream.
+    #[cfg(unix)]
+    fn open_stream(&self) -> StreamHandle {
+        StreamHandle::new(self.verbosity)
```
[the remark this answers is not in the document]
bepri: I don't believe there is an alternative, at least not one that's any easier. If `JoinHandle` allowed me to copy it in any way then I could've gotten away with multiple printers existing, but that limitation was a real show stopper.

PR #443, opened 2026-04-22
fix(printer): ignore broken pipe on output streams

### LINKED ISSUE #378: Broken pipe when trimming lines (head)
```
❯ snapcraft revisions snapcraft | head -n1
stead. The old name will be removed in a future release. Rev.    Uploaded              Arches    Version                   Channels

Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

gcomneno:
## Summary

Ignore `BrokenPipeError` when writing to output streams in the printer.

This avoids crashes when output is piped to an early-closing consumer (for example
`head -n1`) and the downstream stream closes before craft-cli finishes writing or
cleaning up the terminal/output state.

## Testing

- `pytest -q tests/unit/test_printer.py -k "broken_pipe" -rs`
- `pytest -q tests/unit/test_printer.py -rs`
- `pytest -q tests/integration/test_messages_integration.py -k "third_party_output" -rs`
- `pytest -q tests/unit/test_messages_stream_cm.py -rs`
- `ruff check craft_cli/printer.py tests/unit/test_printer.py`

## AI assistance

AI assistance was used while investigating and preparing the review fix; the resulting code and tests were manually verified.

### DIFF
--- a/craft_cli/printer.py
+++ b/craft_cli/printer.py
@@ -29,6 +29,7 @@ import threading
 import time
 import weakref
 from collections.abc import Callable
+from contextlib import suppress
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import lru_cache
@@ -50,6 +51,12 @@ TESTMODE = False
 ANSI_CLEAR_LINE_TO_END = "\x1b[K"  # ANSI escape code to clear the rest of the line.
 ANSI_HIDE_CURSOR = "\x1b[?25l"
 ANSI_SHOW_CURSOR = "\x1b[?25h"
+
+
+def _safe_print(*args: Any, **kwargs: Any) -> None:
+    """Print to a stream, ignoring BrokenPipeError from downstream consumers."""
+    with suppress(BrokenPipeError):
+        print(*args, **kwargs)
 
 
 @dataclass
@@ -233,7 +240,7 @@ class Printer:
         if not TESTMODE:
             self.spinner.start()
             if _supports_ansi_escape_sequences() and _stream_is_terminal(sys.stderr):
-                print(ANSI_HIDE_CURSOR, end="", file=sys.stderr, flush=True)
+                _safe_print(ANSI_HIDE_CURSOR, end="", file=sys.stderr, flush=True)
         weakref.finalize(self, self.stop)
 
     def set_terminal_prefix(self, prefix: str) -> None:
@@ -293,11 +300,11 @@ class Printer:
         ):
             # If the last message's stream is different from this new one,
             # send a carriage return to the original stream only.
-            print("\r", flush=True, file=self.prv_msg.stream, end="")
+            _safe_print("\r", flush=True, file=self.prv_msg.stream, end="")
             previous_line_end = ""
         if self.prv_msg and previous_line_end == "\n":
             previous_line_end = ""
-            print(flush=True, file=self.prv_msg.stream)
+            _safe_print(flush=True, file=self.prv_msg.stream)
 
         # fill with spaces until the very end, on one hand to clear a possible previous message,
         # but also to always have the cursor at the very end
@@ -320,11 +327,11 @@ class Printer:
             line = _format_term_line(
                 previous_line_end, text, spintext, ephemeral=message.ephemeral
             )
-            print(line, end="", flush=True, file=message.stream)
+            _safe_print(line, end="", flush=True, file=message.stream)
 
         if message.end_line:
             # finish the just shown line, as we need a clean terminal for some external thing
-            print(flush=True, file=message.stream)
+            _safe_print(flush=True, file=message.stream)
             self.unfinished_stream = None
         else:
             self.unfinished_stream = message.stream
@@ -340,7 +347,7 @@ class Printer:
         else:
             text = message.text
 
-        print(text, file=message.stream)
+        _safe_print(text, file=message.stream)
 
     def _write_bar_terminal(self, message: _MessageInfo) -> None:
         """Write a progress bar to the screen."""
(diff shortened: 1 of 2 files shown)


gcomneno:
The functional fix looks healthy: all test jobs passed. The remaining failures appear to be external to this change:
- QA is red only in lint/ty on pre-existing diagnostics outside this PR's diff
- Check policy is red in OSV-scanner due to pytest 9.0.2 reported from the generated requirements artifact

gcomneno:
Hi! Just gently checking whether there’s anything I should adjust here.

I rechecked this locally against the current upstream/main and re-ran the targeted checks:

- `uv run --group dev python -m pytest -q tests/unit/test_printer.py -rs`
- `uv run --group lint --group types ruff check craft_cli/printer.py tests/unit/test_printer.py`
- `uv run --group lint --group types ruff format --check craft_cli/printer.py tests/unit/test_printer.py`

The remaining CI failures still look unrelated to this PR’s diff: `ty` reports diagnostics outside `craft_cli/printer.py` and `tests/unit/test_printer.py`, while the other lint/type checks pass.

Happy to rework this if useful, or to close it if this should not move forward.

craft_cli/printer.py
```diff
@@ -356,7 +356,7 @@ def _write_line_captured(self, message: _MessageInfo) -> None:
         else:
             text = message.text
 
-        print(text, file=message.stream)
+        _safe_print(text, file=message.stream)
```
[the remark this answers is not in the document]
gcomneno: Thanks — I reproduced the shutdown case against the current PR head before the fix: with a real `sys.stdout` pipe to `head -n1`, the producer exited with status 120 and emitted `Exception ignored ... BrokenPipeError` during interpreter shutdown.

The follow-up commit `a33a4d6` now forces the captured standard-stream write to flush inside `_safe_print()`. If that raises `BrokenPipeError`, only `sys.stdout`/`sys.stderr` have their underlying descriptor redirected to `os.devnull`; arbitrary streams keep the previous policy and are not redirected.

I also added a subprocess regression test using a real producer/`head -n1` pipeline. The test is RED on the previous head (producer status 120) and GREEN with this fix (status 0, no `BrokenPipeError`).

Local validation is green: 154 printer tests, 25 stream-context tests, 10 related integration tests, plus Ruff check/format.

craft_cli/printer.py
```diff
@@ -56,8 +56,20 @@
 
 def _safe_print(*args: Any, **kwargs: Any) -> None:
     """Print to a stream, ignoring BrokenPipeError from downstream consumers."""
-    with suppress(BrokenPipeError):
+    stream = kwargs.get("file")
+    if stream is None:
+        stream = sys.stdout
```
bepri:
```suggestion
    stream = kwargs.get("file", sys.stdout)
```

tests/unit/test_printer.py
```diff
@@ -1208,6 +1210,74 @@ def test_stop_streams_unfinished_err(capsys, log_filepath):
     assert err == "\n"
 
 
+def test_show_captured_broken_pipe_does_not_raise(log_filepath, monkeypatch):
```
bepri: I think this and the next two tests could be combined into a singular, parametrized test. It should run once for `ephemeral=True`, and once for `ephemeral=False`. Both cases should do a basic `show()` and then a `stop()` with an unfinished stream to demonstrate that all scenarios do not throw

tests/unit/test_printer.py
```diff
@@ -1602,3 +1672,65 @@ def test_unfinished_stream_closed(log_filepath, mocker):
     # unfinished message on the (now closed) stream.
     stream.close()
     printer.stop()
+
+
+def test_captured_stdout_broken_pipe_does_not_fail_at_shutdown():
```
bepri: This is a good test, but it's really an integration test considering all the forking going on. Could you move this to `tests/integration/test_printer.py`?

craft_cli/printer.py
```diff
@@ -56,8 +56,20 @@
...
+
+    try:
         print(*args, **kwargs)
+    except BrokenPipeError:
+        if stream in (sys.stdout, sys.stderr):
+            with suppress(OSError, ValueError):
```
bepri, line 66: Just a little breadcrumb to help future adventurers
```suggestion
						# The pipe is broken, dump all remaining output to /dev/null instead
            with suppress(OSError, ValueError):
```

bepri (commented):
Thanks, this seems to work in local testing! I just have a couple of requests, but the meat of the solution is good :)

craft_cli/printer.py
gcomneno: Updated in `4276d5c`: `_safe_print()` now uses `kwargs.get("file", sys.stdout)` directly.

tests/unit/test_printer.py
gcomneno: Updated in `4276d5c`: the broken-pipe unit coverage is now a single parametrized test for `ephemeral=False` and `ephemeral=True`; both paths exercise `show()` followed by `stop()` with an unfinished stream.

tests/unit/test_printer.py
gcomneno: Moved in `4276d5c` to a new `tests/integration/test_printer.py`. The real producer/`head -n1` regression still verifies producer exit status 0 and no `BrokenPipeError` during shutdown.

craft_cli/printer.py
gcomneno: Added in `4276d5c`, immediately before redirecting the broken standard stream to `/dev/null`.

bepri (approved):
LGTM, thanks :)

lengau (approved):
Thanks!

PR #445, opened 2026-04-23
fix(printer): handle multiline permanent progress correctly

gcomneno:
## Summary

Handle multiline permanent progress updates correctly when replacing a previous temporary progress message.

Previously, a permanent multiline progress message could leave leftover text from the previous temporary progress line on the terminal, and multiline progress text was also logged as a single string containing embedded newlines.

This change formats terminal multiline progress output line-by-line so each rendered line is fully cleared, and logs multiline message text as separate timestamped lines.

## Testing

```bash
pytest -q tests/integration/test_messages_integration.py -k "multiline_permanent_progress_overwrites_temporary_progress" -rs -vv
pytest -q tests/unit/test_printer.py -k "_log or writelineterminal" -rs
ruff check craft_cli/printer.py tests/integration/test_messages_integration.py tests/unit/test_printer.py
```

### DIFF
--- a/craft_cli/printer.py
+++ b/craft_cli/printer.py
@@ -29,6 +29,7 @@ import threading
 import time
 import weakref
 from collections.abc import Callable
+from contextlib import suppress
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


gcomneno:
The functional test matrix is passing. The remaining failures appear unrelated to this change:

- OSV-scanner reports GHSA-6w46-j5rx-g56g for pytest 9.0.2 in generated requirements, while the source scan reports no issues.
- lint / files fails only in ty diagnostics outside this PR's diff. This PR changes only craft_cli/printer.py and tests/integration/test_messages_integration.py; ruff, mypy, pyright, formatting, and the test jobs are passing.

bepri (approved):

lengau (approved):
Thanks!

PR #456, opened 2026-06-29
docs: use folders for each axis

bepri:
- [ ] Have you successfully run `make lint && make test`?

---

This is done in preparation for a Starbase update. I didn't make any attempt to improve content or fix anti-patterns within any of the files, I only aimed to get Craft CLI aligned with the structure of other projects.

- Made Reference, Tutorials, and Explanation into folders with top-level `index.rst` files and each topic moved into its own file
- Added redirects for broken links this may have created
- Fixed a docs build issue that sometimes came up


docs/conf.py
```diff
@@ -50,9 +48,10 @@ def run_apidoc(_):
 
     sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
     cur_dir = os.path.abspath(os.path.dirname(__file__))
+    target = os.path.join(cur_dir, "reference")
     module = os.path.join(cur_dir, "..", "craft_cli")
-    exclude_patterns = ["*pytest_plugin*"]
-    main(["-e", "--no-toc", "--force", "-o", cur_dir, module, *exclude_patterns])
+    exclude_patterns = ["*pytest_plugin*", "*dispatcher*"]
```
bepri, line 53: Everything in `craft_cli.dispatcher` was re-exported in the top-level `craft_cli` module. This meant that for the purposes of autogen docs from apidoc, `CommandGroup` could resolve to either `craft_cli.CommandGroup` or `craft_cli.dispatcher.CommandGroup`, and the ambiguity was raised as an error.

Since everything is re-exported, this just excludes that module from the autogen docs. This isn't a very elegant approach, but these autogen docs need a closer look than what is reasonable to do in this PR so this is the quick and dirty solution.

docs/conf.py
```diff
@@ -14,8 +14,6 @@
     "github_url": "https://github.com/canonical/craft-cli",
 }
 
-linkcheck_ignore = ["craft_cli.dispatcher.html#craft_cli.dispatcher.CommandGroup"]
```
bepri, line 17: This ignore was solving a similar issue caused by the same root problem, and was no longer needed now that it isn't being generated anymore.

docs/explanation/command-groups.rst
bepri, line 1: These were verbatim migrations, I only took the time to fix links.

docs/how-to/set-default-command.rst
medubelko, line 1: Thoughts on shifting these to `how-to/`? I know I originally introduced this path, but since we're aligning the dirs, I feel this would be a good time to do it.

It would require redirecting all the how-to pages.

docs/explanation/index.rst
medubelko, line 1: The title should be "Explanation", singular.

docs/index.rst
```diff
@@ -5,15 +5,15 @@ Welcome to Craft CLI's documentation!
    :maxdepth: 1
    :hidden:
 
-   tutorials
+   tutorials/index
    how-to-guides/index
```
medubelko:
```suggestion
   how-to/index
```

medubelko (changes requested):

medubelko (approved):

jahn-junior (approved):
Did a safety check on the index pages and overall structure, and everything looks good to me. Thanks!

PR #467, opened 2026-08-03
fix(pytest_plugin): reject negative advance in _RecordingProgresser

### LINKED ISSUE #466: bug: `_RecordingProgresser.advance` accepts negatives the real progresser rejects
### Bug Description

noticed while reviewing https://github.com/canonical/craft-application/pull/994 where it silently masked a bug.

the `emitter` fixture swaps the real progress bar for `_RecordingProgresser`, but the mock class does not enforce a constraint the real class does. real `_Progresser.advance` raises on `amount < 0`.

### To Reproduce

```python
from craft_cli import emit

def test_negative_advance(emitter):
    with emit.progress_bar("downloading", 100) as progress:
        progress.advance(-1)  # expected ValueError, got a recorded call
    assert emitter.interactions[-1].args == ("advance", (-1,), {})
```

^ passes, bit shouldn't

GreyViperTooth:
## What

The `emitter` pytest fixture swaps the real progress bar for `_RecordingProgresser`, but that mock did not enforce a constraint the real `_Progresser.advance` does: it raises `ValueError` when `amount < 0`. As a result, tests using the fixture silently accepted negative advances that would crash in production, which can mask real bugs (this was noticed in canonical/craft-application#994).

## How

- Give `_RecordingProgresser.advance` the same single-positional `amount: float` signature as the real `_Progresser.advance`.
- Add the same negative-amount guard (`raise ValueError` on `amount < 0`).
- The recorded-call format is unchanged (`call(advance, 100)`), so existing fixture assertions still hold.
- Add a regression test asserting the fixture now rejects a negative advance.

Fixes #466

🤖 Generated with [Claude Code](https://claude.com/claude-code)


craft_cli/pytest_plugin.py
```diff
@@ -238,9 +238,15 @@ def __enter__(self) -> Self:
...
         return False  # do not consume any exception
 
-    def advance(self, *a: Any, **k: Any) -> None:
-        """Record the advance usage."""
-        self.recording_emitter.record("advance", a, k)
+    def advance(self, amount: float) -> None:
+        """Record the advance usage.
+
+        Mirrors the validation done by the real ``_Progresser.advance`` so that
+        tests using the ``emitter`` fixture don't silently accept invalid usage.
```
bepri: We can skip this addition to the docstring, I think.

craft_cli/pytest_plugin.py
```diff
@@ -238,9 +238,15 @@ def __enter__(self) -> Self:
     def __exit__(self, *_exc_info: object) -> Literal[False]:
         return False  # do not consume any exception
 
-    def advance(self, *a: Any, **k: Any) -> None:
-        """Record the advance usage."""
-        self.recording_emitter.record("advance", a, k)
+    def advance(self, amount: float) -> None:
```
bepri: This breaks the API of this fixture, I would do this instead:


```suggestion
    def advance(self, *a: Any, amount: float, **k: Any) -> None:
```

(and update line 249 accordingly)

bepri (commented):
Thanks! I left a little feedback below. Could you also add this as a bugfix in the changelog at `docs/changelog.rst`?

GreyViperTooth:
Thanks @bepri! Kept the *a/**k passthrough and dropped the docstring as suggested. One tweak: I put amount first (def advance(self, amount, *a, **k)) instead of keyword-only, since *a, amount, **k would break the existing positional call progress_bar.advance(100). Also added the changelog entry.

bepri (approved):

lengau (approved):
Thanks!

